PR #32, opened 2016-12-21
Purge with truncate for Doctrine ORM

vetal2409:
Fixed bug regarding to issue [#292](https://github.com/hautelook/AliceBundle/issues/292).
!!! For Doctrine ORM

### DIFF
--- a/src/Bridge/Doctrine/Purger/OrmPurger.php
+++ b/src/Bridge/Doctrine/Purger/OrmPurger.php
@@ -79,9 +79,12 @@ use Nelmio\Alice\IsAServiceTrait;
 
     /**
      * @inheritdoc
+     * @throws \Doctrine\DBAL\DBALException
      */
     public function purge()
     {
+        $this->manager->getConnection()->prepare('SET FOREIGN_KEY_CHECKS = 0;')->execute();
         $this->purger->purge();
+        $this->manager->getConnection()->prepare('SET FOREIGN_KEY_CHECKS = 1;')->execute();
     }
 }
--- a/src/Loader/PurgerLoader.php
+++ b/src/Loader/PurgerLoader.php
@@ -55,13 +55,23 @@ use Nelmio\Alice\IsAServiceTrait;
      *
      * {@inheritdoc}
      */
-    public function load(array $fixturesFiles, array $parameters = [], array $objects = [], PurgeMode $purgeMode = null): array
-    {
-        if (null === $purgeMode) {
-            $purgeMode = PurgeMode::createDeleteMode();
+    public function load(
+        array $fixturesFiles,
+        array $parameters = [],
+        array $objects = [],
+        PurgeMode $purgeMode = null
+    ): array {
+        if (null !== $purgeMode) {
+            $purger = $this->purgerFactory->create($purgeMode);
+        } else {
+            if ($this->purgerFactory instanceof PurgerFactoryInterface) {
+                $purger = $this->purgerFactory;
+            } else {
+                $purgeMode = PurgeMode::createDeleteMode();
+                $purger = $this->purgerFactory->create($purgeMode);
+            }
         }
 
-        $purger = $this->purgerFactory->create($purgeMode);
         $purger->purge();
 
         return $this->loader->load($fixturesFiles, $parameters, $objects);


src/Bridge/Doctrine/Purger/OrmPurger.php
```diff
@@ -79,9 +79,12 @@ public function create(PurgeMode $mode, PurgerInterface $purger = null): PurgerI
 
     /**
      * @inheritdoc
+     * @throws \Doctrine\DBAL\DBALException
      */
     public function purge()
     {
+        $this->manager->getConnection()->prepare('SET FOREIGN_KEY_CHECKS = 0;')->execute();
```
theofidry: Is this necessarily? This happen only because the cascade delete or so are not set properly. Using this is a shortcut I'm using as well sometimes but I'm not sure if it should be done in the core

src/Bridge/Doctrine/Purger/OrmPurger.php
```diff
@@ -79,9 +79,12 @@ public function create(PurgeMode $mode, PurgerInterface $purger = null): PurgerI
 
     /**
      * @inheritdoc
+     * @throws \Doctrine\DBAL\DBALException
```
theofidry:
```php
/**
  * {@inheritdoc}
  *
  * @throws \Doctrine\DBAL\DBALException
  */
```

src/Bridge/Doctrine/Purger/OrmPurger.php
vetal2409: But I don't need cascade delete, therefore I can't execute command with  --purge-with-truncate option because of foreign keys... How I can do this?

vetal2409:
$purgeMode is defined in DoctrineOrmLoader regarding to the option but then, somehow,  it doesn't pass it to PurgeLoader:load as 4th parameter. So that I have added condition to check if it doesn't passed then it should take it from property.  I know that it's not good way but I don't know if there are other places where it calls or will call this PurgeLoader:load  method.

src/Bridge/Doctrine/Purger/OrmPurger.php
WillGibson: I think the issue comes from the truncating of the tables not necessarily happening in a good order for cascade delete to work, and I guess some people may not want cascade delete for some reason, so I would be in favour of keeping this in.

src/Loader/PurgerLoader.php
```diff
@@ -55,13 +55,23 @@ public function __construct(
...
+    ): array {
+        if (null !== $purgeMode) {
+            $purger = $this->purgerFactory->create($purgeMode);
+        } else {
+            if ($this->purgerFactory instanceof PurgerFactoryInterface) {
+                $purger = $this->purgerFactory;
+            } else {
+                $purgeMode = PurgeMode::createDeleteMode();
+                $purger = $this->purgerFactory->create($purgeMode);
+            }
```
WillGibson: I'm not sure this change is required, line 67 doesn't look right to me anyway.

src/Bridge/Doctrine/Purger/OrmPurger.php
WillGibson: Oh, but actually, this is probably MySQL specific, so I don't think it should be here in case someone tries to use it with another kind of database, maybe setting up cascade delete properly will work instead. I'll have a go at that on our project and report back.

theofidry: No indeed, this should be done in a MySQL purger or check the DB platform first

src/Loader/PurgerLoader.php
theofidry: yeah, I think we can just keep L70-71 of the L67-L72 block

WillGibson:
Yes, I'll have a look later, probably I'll just make a new PR because I
guess I can't work on this branch directly?

theofidry:
Yeah, do as it's more convenient for you :)

theofidry:
Truncate should work properly once https://github.com/hautelook/AliceBundle/issues/366 is fixed which hopefully is very soon. Thanks again for looking into this :)

One may still need the foreign key checks disabled, but I would prefer to see if that is really necessary later on

robertfausk:
Foreign key checks needs to be disabled. On current dev-master I'm getting following error:

```
bin/console hautelook:fixtures:load --no-interaction --purge-with-truncate  
07:13:22 ERROR     [console] Error thrown while running command "hautelook:fixtures:load --no-interaction --purge-with-truncate". Message: "An exception occurred while executing 'TRUNCATE multiple_choice_answer':

SQLSTATE[42000]: Syntax error or access violation: 1701 Cannot truncate a table referenced in a foreign key constraint ...
```

PR #33, opened 2017-01-11
Allow using different names than "default" for entity manager.

robotzero:
I would like to have an ability to use different names than "default" for my entity managers when using AliceDataFixtures. Doctrine has default_entity_manager option and it will point specified entity manager name to orm_entity_manager in a container or will use first on the list of configured entity managers.


theofidry:
Hm right, I thought `default_entity_manager` was the alias for `entity_manager` and not vice versa when I wrote that piece of code.

Thanks for fixing it :)

Note that I may not do a release right away: I'm a bit busy right now, so if it's urgent, I recommend to use the `@dev` version or the specific commit.

PR #38, opened 2017-02-27
Add support for doctrine ODMs

alekitto:
Add support for Doctrine PHPCR and MongoDB ODMs.
Persister does not need any modification as it works with the common `ObjectManager` interface.

TODO:
 - [x] integration tests


vendor-bin/doctrine/composer.json
```diff
@@ -8,8 +8,13 @@
         }
     },
     "require-dev": {
+        "alcaeus/mongo-php-adapter": "^1.0",
```
theofidry: I would prefer have dedicated (barmani) namespaces here

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
```diff
@@ -20,7 +20,7 @@
         </service>
 
         <service id="fidry_alice_data_fixtures.persistence.purger.doctrine.orm_purger"
```
theofidry: the service name needs to be updated as well

src/Bridge/Symfony/DependencyInjection/FidryAliceDataFixturesExtension.php
```diff
@@ -44,12 +41,12 @@ public function load(array $configs, ContainerBuilder $container)
         $processedConfiguration = $this->processConfiguration($configuration, $configs);
         $bundles = array_flip($container->getParameter('kernel.bundles'));
 
-        if (false === array_key_exists(self::NELMIO_ALICE_BUNDLE_CLASS, $bundles)) {
+        if (false === array_key_exists(NelmioAliceBundle::class, $bundles)) {
```
theofidry: doesn't this break tests? If the bundle is not installed then this check fails which defeat its point, hence the usage of constants here

src/Bridge/Doctrine/Purger/Purger.php
```diff
@@ -0,0 +1,115 @@
...
+    public function purge()
+    {
+        $this->purger->purge();
+    }
+
+    private static function createPurger(ObjectManager $manager): DoctrinePurgerInterface
+    {
+        if ($manager instanceof EntityManagerInterface) {
+            return new DoctrineOrmPurger($manager);
+        } elseif ($manager instanceof DoctrinePhpCrDocumentManager) {
```
theofidry: no need for an `else` as you do a return

src/Bridge/Doctrine/Purger/Purger.php
```diff
@@ -0,0 +1,115 @@
...
+use Fidry\AliceDataFixtures\Persistence\PurgerFactoryInterface;
+use Fidry\AliceDataFixtures\Persistence\PurgerInterface;
+use Nelmio\Alice\IsAServiceTrait;
+
+/**
+ * Bridge for Doctrine purger.
+ *
+ * @author Vincent CHALAMON <vincentchalamon@gmail.com>
+ */
+class Purger implements PurgerInterface, PurgerFactoryInterface
```
theofidry: the comments regarding the `final` are missing

src/Bridge/Symfony/DependencyInjection/FidryAliceDataFixturesExtension.php
alekitto: The check is the same, as the FQCN resolution happens at compile time.
So if the class does not exists it doesn't break.

However, this modification was pushed by error, as it wasn't intended to be in this PR, I'll revert it if you prefer.

theofidry: >The check is the same, as the FQCN resolution happens at compile time.
So if the class does not exists it doesn't break.

Not sure to understand, if `Foo` doesn't exist using `Foo::class` will throw an error.

alekitto: > Not sure to understand, if Foo doesn't exist using Foo::class will throw an error.

Nope. PHP converts to string every use of `::class` constants at compile time as documented here: http://php.net/manual/en/language.oop5.constants.php.
So classes don't need to be loaded nor exist before using the constant.
See here for an example: https://3v4l.org/sHMTh

theofidry: Hm maybe the problem I stumbled on was a different one then. You can leave this change in the PR it's not much and well scoped. Cool to simplify a thing or two :)

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
alekitto: I have to create two additional services for the ODMs as well, but i think this service can be left as is, as `orm` is present in its name.
The addtional services should be `fidry_alice_data_fixtures.persistence.purger.doctrine_mongodb.purger` and `fidry_alice_data_fixtures.persistence.purger.doctrine_phpcr.purger`.

Is it ok for you?

vendor-bin/doctrine_mongodb/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "autoload": {
```
theofidry: What an horrible thing to do 😱 !

More seriously, it's a bit tedious to do that and error prone, that's what `theofidry/composer-inheritance-plugin` is for so you shouldn't need this.

alekitto: I'm not sure to understand what you mean.
However, i've just replicated the `composer.json` in `vendor-bin/doctrine` folder and adapted the requirements, i thought it was ok...

theofidry: What `theofidry/composer-inheritance-plugin` does is actually merging a part of the config of the parent `composer.json` into this one.

For example if you check [this file](https://github.com/nelmio/alice/blob/master/vendor-bin/symfony/composer.json), you'll see that there is only that plugin and Symfony... because it inherits the whole autoloading and required packages config from the root one.

If you want more details on it, I suggest you to dig in https://github.com/theofidry/composer-inheritance-plugin, it's actually not much :)

alekitto: Sure! Got it. I didn't see that the same namespaces were present in the main `composer.json` file. I'll remove those blocks from the children.

alekitto:
Like you said, the integration tests were the hardest part, but should be all done now.

theofidry:
Nice!

I'll take a proper look at it over the weekend, I'll try to merge/fix a few things and to a new release for the whole alice ecosystem.

alekitto:
@theofidry Any news on this?

fixtures/Bridge/Symfony/SymfonyApp/DoctrineMongodbKernel.php
```diff
@@ -0,0 +1,49 @@
...
+    public function registerBundles()
+    {
+        $bundles = [
+            new FrameworkBundle(),
+            new NelmioAliceBundle(),
+            new FidryAliceDataFixturesBundle(),
+            new DoctrineMongoDBBundle(),
+        ];
+
+        if (class_exists('Fidry\PsyshBundle\PsyshBundle')) {
```
theofidry: this could be changed to `\Fidry\PsyshBundle\PsyshBundle::class`

fixtures/Bridge/Symfony/SymfonyApp/DoctrinePhpcrKernel.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $bundles = [
+            new FrameworkBundle(),
+            new NelmioAliceBundle(),
+            new FidryAliceDataFixturesBundle(),
+            new DoctrineBundle(),
+            new DoctrinePHPCRBundle(),
+        ];
+
+        if (class_exists('Fidry\PsyshBundle\PsyshBundle')) {
```
theofidry: same as above

composer.json
```diff
@@ -13,11 +13,12 @@
     ],
     "require": {
         "php": "~7.0.0 || ~7.1.0",
-        "nelmio/alice": "^3.0@beta"
+        "nelmio/alice": "^3.0-beta.2@beta"
```
theofidry: could you upgrade to `.5`?

composer.json
```diff
@@ -13,11 +13,12 @@
...
     },
     "require-dev": {
         "bamarni/composer-bin-plugin": "^1.0",
         "phpunit/phpunit": "^5.5",
+        "phpspec/prophecy": "^1.6",
```
theofidry: is this necessary?

alekitto: Yes, to be compatible with PHP7 in `--prefer-lowest` tests.

alekitto:
I think that packagist did not updated the `nelmio/alice` package for beta.5 version. Composer fails with this error: `The requested package nelmio/alice ^3.0-beta.5@beta is satisfiable by nelmio/alice[3.0.x-dev] but these conflict with your requirements or minimum-stability.`

src/Bridge/Symfony/DependencyInjection/FidryAliceDataFixturesExtension.php
```diff
@@ -44,66 +43,51 @@ public function load(array $configs, ContainerBuilder $container)
...
-        }
-
-        if ($doctrineBundleIsRegistered) {
-            $loader->load('doctrine_orm.xml');
-        }
-    }
-
-    private function registerEloquentConfig(array $bundles, array $configs, LoaderInterface $loader)
-    {
+    private function registerConfig(
```
theofidry: a quick phpdoc is missing here as well to specify what `$bundles` and `$configs` are

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
theofidry: Yeah I'm cool with that (and they should be in their respective `.xml` files

theofidry:
The message error is a bit weird... Let's wait a bit and see

theofidry:
Indeed it is missing here: https://packagist.org/packages/nelmio/alice

Let me ping Jordi to update packagist as I don't have any access there

theofidry:
@alekitto should be good now, although it looks like the Eloquent bridge is broken.

alekitto:
Rebased and squashed.

bin/test.sh
```diff
@@ -37,6 +47,14 @@ vendor-bin/doctrine/bin/doctrine o:s:c
 
 $PHPUNI_PREFIX vendor-bin/doctrine/bin/phpunit -c phpunit_doctrine.xml.dist $PHPUNIT_FLAGS
```
theofidry: there's a typo: `$PHPUNIT_PREFIX`

vendor-bin/eloquent/composer.json
```diff
@@ -5,8 +5,8 @@
         ]
     },
     "require-dev": {
-        "illuminate/database": "~5.3.0",
-        "illuminate/filesystem": "5.0 - 5.3",
+        "illuminate/database": "^5.3",
```
theofidry: Illuminate is not semver, so a `~` is required as you are allowing updates with BC breaks otherwise

vendor-bin/symfony/composer.json
```diff
@@ -5,9 +5,16 @@
...
+        "alcaeus/mongo-php-adapter": "^1.0.10",
         "doctrine/data-fixtures": "^1.2",
         "doctrine/doctrine-bundle": "^1.6",
+        "doctrine/mongodb": "^1.4",
+        "doctrine/mongodb-odm-bundle": "^3.0",
         "doctrine/orm": "^2.5",
+        "doctrine/phpcr-bundle": "^1.0",
+        "doctrine/phpcr-odm": "^1.4",
+        "jackalope/jackalope-doctrine-dbal": "^1.2",
+        "psy/psysh": "^0.6",
```
theofidry: I don't think we require `psy/psysh` and I would prefer to avoid it: it brings a lot of undesirable dependencies

vendor-bin/doctrine_phpcr/composer.json
```diff
@@ -0,0 +1,14 @@
+{
+    "require-dev": {
+        "doctrine/data-fixtures": "^1.2",
+        "doctrine/dbal": "^2.5",
+        "doctrine/phpcr-odm": "^1.4",
+        "jackalope/jackalope-doctrine-dbal": "^1.2",
+        "symfony/console": "^2.5",
```
theofidry: why is the `symfony/console` required here?

vendor-bin/symfony/composer.json
alekitto: You want to remove the psysh-bundle package also?
Otherwise `0.6` is the minimum required version to run the test suite correctly, as its `renderException` method is incompatible with symfony console `Application` class

vendor-bin/doctrine_phpcr/composer.json
alekitto: Is required to run the testsuite as the `cli-config.php` used to initialize PHPCR db schema requires console `HelperSet` class to be available.

vendor-bin/symfony/composer.json
alekitto: In any case, the `psysh-bundle` deletion doesn't break the testsuite

alekitto:
Thinkin' about this: in some of my projects i've used the ORM and one or both the ODMs together. Why not to create a `CompositePersister` where all the doctrine persisters are subsequently called to load the fixtures in a one-shot command?
This can be also used to load fixtures in ORM-only applications where multiple entity manager are used.

But probably is out of the scope of this PR...

vendor-bin/symfony/composer.json
theofidry: Yeah. It brings too many deps so the difference between the deps the consumer will get and ours during tests will be too great

vendor-bin/doctrine_phpcr/composer.json
theofidry: Damn, fine

theofidry:
The only issue I see with your `CompositePersister` is that it brings the question of which persister should be used and when? Hence you could complexify the `Persister` interface to have a `supports()` method and then have a chain which calls the first persister supporting the object to be persisted.

It is a fine idea, but it brings significant complexity. I do see the point in some applications, but I'm not sure if that should be included in the core. In any case this should not be done in this PR :)

alekitto:
I'll open a new issue when this PR will be merged, to discuss it further.

theofidry:
Thanks very much for that one :)

omarsoufiane:
can you add some instructions on the documentation? i ave some problems setting up alice with mongodb, and i'm not sure if the config db-driver: doctrine orm is supposed to work

PR #42, opened 2017-05-22
Refactor loader

theofidry:
Related to https://github.com/theofidry/AliceDataFixtures/pull/32.

I think I originally didn't wanted to pollute the `LoaderInterface` with `PurgeMode` which is specific to only one loader, but this causes two problems:

- loaders decorate each others, so this needs to be passed around
- as it's not part of the interface it can be omitted by the consumers

Given that this library `LoaderInterface` is mostly about persistence, I think it's ok to include it there. 

/cc @WillGibson @vetal2409


theofidry:
Erf, don't understand why it complains about a trait supposed to be here since 3.0.0-beta.5...

WillGibson:
I'm getting...

```
  [Symfony\Component\Debug\Exception\ClassNotFoundException]                                            
  Attempted to load class "OrmPurger" from namespace "Fidry\AliceDataFixtures\Bridge\Doctrine\Purger".  
  Did you forget a "use" statement for another namespace? 
```
Do this line in \Hautelook\AliceBundle\Loader\DoctrineOrmLoader::loadFixtures()...
```
$purger = new OrmPurger($manager, $purgeMode);
```
...needed to be...
```
$purger = new Purger($manager, $purgeMode);
```

I changed this locally and that error went away, but I'm still getting...
```
SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails
```
...if I try to use the --purge-with-truncate option.

theofidry:
I think HautelookAliceBundle needs to be updated accordingly for this to work

WillGibson:
Thanks @theofidry 

I ran this to try and update anything that might be related to this as I can see a load of newer releases talking about fixes relating to purging in various Doctrine related repositories...

```composer update hautelook/alice-bundle nelmio/alice theofidry/alice-data-fixtures doctrine/annotations doctrine/cache doctrine/collections doctrine/common doctrine/data-fixtures doctrine/dbal doctrine/doctrine-bundle doctrine/doctrine-cache-bundle doctrine/doctrine-fixtures-bundle doctrine/doctrine-migrations-bundle doctrine/inflector doctrine/instantiator doctrine/lexer doctrine/migrations doctrine/orm 
```

I was still getting a "bork" when running the command with --purge-with-truncate, but I think it might be an issue with one of our foreign keys, if I get rid of it all works well. I'll go back to our entities and see if we configured them wrong somehow.

theofidry:
I don't think Doctrine changed anything as of late for this. I'll look into it, but I think it's just HautelookAliceBundle that needs to be updated (I mean update the code, not you your dependencies)

oscarnevarezleal:
Hi @theofidry , So what would it be the path to follow in order to get HautelookAliceBundle to work with this project? Should I downgrade HautelookAliceBundle or AliceDataFixtures ?

How can I help?

theofidry:
Well, there's three ways to go:

- Try to update HautelookAliceBundle to this PR and see if that solves the issue (either by relying on a [VCS repo](https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository) or a symlink (i.e. manually put the code in the vendor instead of relying on packagist)
- Waiting for me to free a window to do it myself
- Go backward and downgrade

Thanks for taking a look, that's appreciated :)

WillGibson:
@theofidry I know you have a couple of fails in the build, but I've just tested this locally on a branch of one of our projects and I'm happy to report that it seems to be behaving perfectly :-)

theofidry:
glad to hear :) I now "just" need to get all those builds green which is kinda a pain, especially for this library, but we'll get there eventually

theofidry:
I don't know what's the issue with the lowest dependencies really, will try to fix that later...

WillGibson:
Awesome happy merge :-) THanks @theofidry.

oscarnevarezleal:
Thank you @theofidry

PR #62, opened 2017-10-29
Add documentation for auto primary key doctrine generators

### LINKED ISSUE #40: [Doctrine] Support for disabling ID generators if ID provided in fixture
If you have a Doctrine custom PK generator like [ramsey/uuid-doctrine](https://github.com/ramsey/uuid-doctrine), you can't set that auto-PK from fixture because [it's getting overwritten by Doctrine](http://stackoverflow.com/questions/5301285/explicitly-set-id-with-doctrine-when-using-auto-strategy) after you've set it.

If provided in fixture, It would be nice to disable Doctrine auto-PK generator and use the one provided from fixture.

theofidry:
Closes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ AliceDataFixtures
 1. [Basic usage](#basic-usage)
 1. [Processors](#processors)
 1. [FAQ](#faq)
+    1. [(Doctrine) My IDs are overridden, can I set them?](#doctrine-my-ids-are-overridden-can-i-set-them)
 1. [Contributing](#contributing)
 
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ AliceDataFixtures
         1. [Configuration](#configuration)
 1. [Basic usage](#basic-usage)
 1. [Processors](#processors)
+1. [FAQ](#faq)
 1. [Contributing](#contributing)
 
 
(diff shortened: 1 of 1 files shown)


umpirsky:
Is it possible to override ID generator for all entities at once (instead of doing it on per entity basis)?

theofidry:
As it's done automatically, if you set the ID for all entities then the ID generator will be overridden for all

umpirsky:
I don't want to set ID, but ID generator, like:

```php
$dummyMetadata = $this->manager->getMetadataFactory()->getMetadataFor(Dummy::class);

// Disable the auto generation of the foreign key
$dummyMetadata->setIdGeneratorType(ClassMetadata::GENERATOR_TYPE_NONE);
$dummyMetadata->setIdGenerator(new MyIdGenerator());
```

But not only for `Dummy::class`, I want to do that for all entities.

theofidry:
oh, yeah not out of the box. But you should be able to easily overwrite the existing persister loader to do that. Maybe it can be refactored a bit to make it easier

theofidry:
Actually overriding https://github.com/theofidry/AliceDataFixtures/blob/master/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php seems more correct. You can override it by creating your own class and registering a service with the same ID.

`@final` tags are here to say this was not intended to be extended. But it can be removed if the use case is valid, I just do this by default to ease the refactoring.

umpirsky:
Thanks @theofidry!

Do you think that making `ORMAssignedGenerator` a service and injecting it into `ObjectManagerPersister` would make sense? It would ease overriding. I can submit PR if you are willing to merge such contrib?

umpirsky:
@theofidry [This](https://github.com/hautelook/AliceBundle/blob/master/src/Loader/DoctrineOrmLoader.php#L169-L177) makes it even harder to override.

theofidry:
>Do you think that making ORMAssignedGenerator a service and injecting it into ObjectManagerPersister would make sense? It would ease overriding. I can submit PR if you are willing to merge such contrib?

that or making the persister extendable and putting some logic in a method to make it easy to override.  I'm fine either way.

>This makes it even harder to override.

Indeed this should be in a factory instead :/

umpirsky:
Fixed in https://github.com/theofidry/AliceDataFixtures/pull/91. Thanks!

PR #67, opened 2017-10-31
Make some loaders persister aware

theofidry:
See https://github.com/hautelook/AliceBundle/pull/371


morticue:
Since I updated to version `v1.0.0-rc.1` of this library `theofidry/alice-data-fixtures`, and it updated `hautelook/alice-bundle` to `v2.0.0-beta.5` the purging before loading the fixtures seems to simply have stopped occurring. All data seems to be appended, resulting in duplicate entries after the load command, `hautelook:fixtures:load`. 

Going back to `v1.0.0-rc.0` in `theofidry/alice-data-fixtures` and the purging works again.

I am not sure what causes this problem, but my current guess is that this PR and it's sister PR-referenced here in the description has something to do with it.

theofidry:
Indeed I found an issue. I'll fix that tonight

PR #68, opened 2017-11-02
[Doctrine] Support for disabling ID generators if ID provided in fixture

### LINKED ISSUE #40: [Doctrine] Support for disabling ID generators if ID provided in fixture

dkarlovi:
Closes #40.

### DIFF
--- a/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
+++ b/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
@@ -13,9 +13,11 @@ declare(strict_types=1);
 
 namespace Fidry\AliceDataFixtures\Bridge\Doctrine\Persister;
 
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 use Doctrine\Common\Persistence\ObjectManager;
-use Doctrine\ORM\Id\AssignedGenerator;
-use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadataInfo as ODMClassMetadataInfo;
+use Doctrine\ORM\Id\AssignedGenerator as ORMAssignedGenerator;
+use Doctrine\ORM\Mapping\ClassMetadataInfo as ORMClassMetadataInfo;
 use Fidry\AliceDataFixtures\Persistence\PersisterInterface;
 use Nelmio\Alice\IsAServiceTrait;
 
@@ -34,7 +36,7 @@ use Nelmio\Alice\IsAServiceTrait;
     private $persistableClasses;
 
     /**
-     * @var ClassMetadataInfo[] Entity metadata, FQCN being the key
+     * @var ClassMetadata[] Entity metadata, FQCN being the key
      */
     private $metadata = [];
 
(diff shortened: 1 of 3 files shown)


src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -74,4 +99,14 @@ private function getPersistableClasses(ObjectManager $manager): array
 
         return $persistableClasses;
     }
+
+    private function getMetadata(string $class): ClassMetadataInfo
```
dkarlovi: It's failing for MongoDB, wrong type.

We can return `Doctrine\Common\Persistence\Mapping\ClassMetadata` here (which is correct), but that doesn't guarantee us `$metadata->idGenerator` which we use to fetch the current generator (to be able to reset it afterward).

theofidry: Hm how does it works for MongoDB then? Isn't there the concept of generator there still with the same issue?

dkarlovi: It should work, but we can't link against the interface because it doesn't provide access to the current generator.

theofidry:
Let's see how that goes

dkarlovi:
I'd rather we didn't have to type-check the metadata instance, but I don't see what can be done here. LGTM, merge? :+1:

theofidry:
Many thanks :)

>BTW had no idea you can push into my fork by default.

You can now when you submit a pull request you have a "allow edit from maintainers". I must say this is quite useful: it avoids to have to do another PR against either the fork or the repo

dkarlovi:
> allow edit from maintainers

I thought that means the PR, like title or description. I should RTFM more. :sob: 

Thanks for the merge. Would you consider a release?

theofidry:
Yep done :)

dkarlovi:
It works in my app, TYVM @theofidry!

Nijusan:
hm, used this in a current application setup and is not working anymore:
https://github.com/Nijusan/alice-phpunit-fixtures

dkarlovi:
@Nijusan it works fine for me (I've just tested it), what version are you using?

Nijusan:
i am working with a symfony v4.0.4 setup with alice bundle
and
"theofidry/alice-data-fixtures": {
        "version": "v1.0.1"
    },

Nijusan:
if i disable this part https://github.com/theofidry/AliceDataFixtures/commit/01738c41d5e205641170baf7cf8294beb1ef9169#diff-f959e9bc14bc7c58d57edb90b8794605R85
it works again.

dkarlovi:
Probably depends on if the generator is pre or post insert. Maybe open a PR which solves this issue for you it can be discussed there?

Nijusan:
ok will do so. but thought this issue is doctrine labeled and the doctrine id generator seems to be always a post insert generator ? so i am wondering how this is working for u with doctrine ?

dkarlovi:
I use Ramsey UUID Doctrine which assigns the ID pre-insert. I'm guessing post-insert wouldn't work, it might not be a good idea to re-assign the generator back in that case.

Nijusan:
thanks for ur help. i think it depends, if not all fixtures of that class are configured with setting the id it would also break when the generator isn re-assign i would think?

so there should at least be a check if it is a post insert generator and if so shoud not re assign, but should check for other objects of that type and if the manual setted id is missing and then printing a warning? i am trying to do a pr and it can be discussed there further... thx

Toflar:
I guess it could be solved by moving the logic from the `ObjectManager` to a `Processor` and disabling uuid during `preProcess`. Wdyt @dkarlovi?

Toflar:
Just tried this and it does not work because the `Processor` is also just one object after the other. It would need to be done even before and then it works 😄 I would need to use a special loader then, or have a new interface for a processor to process all objects even earlier. Like so:

```diff
        $objects = $this->loader->load($fixturesFiles, $parameters, $objects, $purgeMode);
+        $this->logger->info('Pre-pre-processing objects.');
-        $this->logger->info('Pre-processing objects.');

+        foreach ($objects as $id => $object) {
+            foreach ($this->processors as $processor) {
+                if ($processor instanceof WhatTheHeckDoWeNameThis) {
+                  $processor->whatTheHeckDoWeNameThis($id, $object);
+                }
+            }
+        }
+
+        $this->logger->info('Pre-processing objects.');

        foreach ($objects as $id => $object) {
            foreach ($this->processors as $processor) {
                $processor->preProcess($id, $object);
            }

            $this->persister->persist($object);
        }
```

Ping @theofidry

PR #69, opened 2017-11-04
Allow Symfony 4.0

theofidry:
Waiting on some third-party deps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,13 @@ matrix:
       env: SYMFONY_VERSION='~3.3.0'
     - php: '7.1'
       env: SYMFONY_VERSION='~3.4.0@dev'
+    - php: '7.1'
+      env: SYMFONY_VERSION='~4.0.0@dev'
 
   allow_failures:
     - php: nightly
     - env: SYMFONY_VERSION='~3.4.0@dev'
+    - env: SYMFONY_VERSION='~4.0.0@dev'
 
 before_install:
   - set -eo pipefail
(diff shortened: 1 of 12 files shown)


robertfausk:
Which deps are still missing?

theofidry:
Right now WouterJEloquentBundle: https://travis-ci.org/theofidry/AliceDataFixtures/jobs/297308563

robertfausk:
Created an issue for a new tag https://github.com/wouterj/WouterJEloquentBundle/issues/73
cause its dev-master already supports Symfony 4.0

robertfausk:
WouterJ just released a new tag:)
https://github.com/wouterj/WouterJEloquentBundle/releases/tag/1.0.0-RC1

I fear that some other deps will pop up with a fail after update this one.

wouterj:
Please note that I only agree support for the latest version of each major release. So if you really want 3.1 and 3.2 support, you might need to help maintaining my bundle :)

theofidry:
Yeah, you can see there is the same issue with the other libraries. @wouterj I don't mind but I'm quite busy so a bit unrealistic, I won't commit to maintain all the libraries used in the bridges, neither Eloquent or Doctrine. I just provide the bridge, if it doesn't work for the versions people want then they will have to fix it. I think I'll just skip some tests depending of the deps to handle those situations

theofidry:
Blocked by [phpcr-odm](https://packagist.org/packages/doctrine/phpcr-odm) which requires `symfony/console 3.x`

PR #70, opened 2017-11-07
Made persister/purger loaders services public

ogizanagi:
This allows to easily choose between a purger or simple persister loader from tests or doctrine fixtures by accessing them directly from the container.

But in fact, what is bothering me is that the already public `fidry_alice_data_fixtures.loader.doctrine` alias (and similar ones for other ORM/ODMs) is aliasing the purger loader, and there is no way to not purge, [even by passing `null` as `$purgeMode`](https://github.com/theofidry/AliceDataFixtures/blob/master/src/Loader/PurgerLoader.php#L67-L69). 😕


ogizanagi:
Alright, see https://github.com/theofidry/AliceDataFixtures/pull/71 for the `NO_PURGE` mode.

> If we go for the second solution though, I don't think this PR makes sense anymore

Agreed :)

theofidry:
Cheers, I'm closing this one in favour of #71 then

PR #71, opened 2017-11-07
Add a NO_PURGE mode

ogizanagi:
Trying to keep things simple: the `NO_PURGE` mode is handled in `PurgerLoader`, not in each `PurgerInterface` implementation.


theofidry:
Many thanks :)

PR #73, opened 2017-11-25
Disable foreign key checks with truncating with MySQL

### LINKED ISSUE #63: Allow to disable foreign keys
From #32, we could add it to the Purge in the `purge()` method to get access to the connection and disable the foreign key checks before running the actual purge.

As this may not be the desirable default behaviour, we could add an option in the config passed to the purger factory.

/cc @robertfausk

theofidry:
Closes #63.


theofidry:
You have the [option `default_purge_mode`](https://github.com/theofidry/AliceDataFixtures#configuration) to pick the one you want and in the case of MySQL FK checks are disabled automatically.

When Doctrine DataFixtures will ship that in their purge we will be able to remove it here, but in the meantime it's done here.

PR #79, opened 2017-12-11
If $logger is optional, $processors should be too

olvlvl:
If `$logger` is optional, `$processors` should be too. I also added the missing doc for `$logger`.


theofidry:
Thanks :)

PR #80, opened 2017-12-11
These methods are not supposed to have return values

olvlvl:
From what I gathered inspecting the code base, these methods have no return value.

### DIFF
--- a/src/ProcessorInterface.php
+++ b/src/ProcessorInterface.php
@@ -28,18 +28,14 @@ interface ProcessorInterface
      *
      * @param string $id Fixture ID
      * @param object $object
-     *
-     * @return
      */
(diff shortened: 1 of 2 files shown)


theofidry:
this must have been from a generated phpdoc or copy/paste when I ported that piece of code. As we are in 7.1 could you add the void return type?

olvlvl:
Yep! Will do.

olvlvl:
@theofidry Can you check now?

theofidry:
I would just remove the `@return`, it doesn't provide any value :)

olvlvl:
Here you go :)

theofidry:
Many thanks :)

PR #88, opened 2018-02-08
Fix documentation of ProcessorInterface

z38:
This PR fixes the documentation of `ProcessorInterface::{pre,post}Process` which got a second parameter some time ago.


theofidry:
Thanks!

PR #89, opened 2018-02-16
[Doctrine] Support for disabling ID generators if ID provided in fixture - part 2

Nijusan:
the changes in https://github.com/theofidry/AliceDataFixtures/pull/68 are not working properly with the default doctrine identity generator.

the test only passes due to wrong handling with the entity object, it does gets changed from the post insert generator so the assert afterwards is ok.
cloning the original entity du be saved and asserting this one with the database result will break the test, so further changes are required


Nijusan:
ok, the disabling of id generators for custom ids does now also work with post insert generators as default identity generator of doctrine is.
but in this case its fully needed that every defined class in the corresponding fixture has a custom id set, because mixing is not possible. this is also tested + a special exception with information if this is done wrong

dkarlovi:
OK, so it's all or nothing? Probably good enough, IMO.

Nijusan:
Hm, maybe it could be possible to detect that the ID value is empty but the AssignValueGenerator is set so we switch to the doctrine uuid generator and switch back if the value is set for another object and do on? then mixing would be possible... i ll try something if this is prefered over the solution yet

Nijusan:
ok, i think it could be possible, but maybe it's not a good solution at all. it could lead to many edge cases and not fullfill the easy ? 

so switching to another generator if the idValue is empty and no idGenerator is set must implement a check on the given type for the idValue, because just switch to the uuid generator leads to errors if the type of the idValue is an integer. sure its possible to check that to decide wehter to use the uuid generator or not. but what should we use if it is an integer value? we cannot use one of the other default generators of doctrine... so we could create a custom one, trying to generate unique integer values? but is that safe enough?

so maybe it is not a good idea to mix the identity strategy with custom and auto generated values because this could lead to already used primary keys...
so if u want to use custom id's u should be responsible to set correct custom ids for all of your objects... 

any other opinion appreciated! :)

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -78,6 +81,8 @@ public function testIsNotClonable()
      */
     public function testCanPersistAnEntity($entity, bool $exact = false)
     {
+        $originalEntity = clone($entity);
```
theofidry: `clone $entity`

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -88,10 +93,41 @@ public function testCanPersistAnEntity($entity, bool $exact = false)
...
         }
     }
 
+    public function testCanPersistMultipleEntitiesWithExplicitIdentifierSet()
+    {
+        $dummy = new DummyWithIdentifier();
+        $dummy->id = 100;
+        $this->persister->persist($dummy);
+
+        $dummy = new DummyWithIdentifier();
```
theofidry: what is the difference with the previous one? That this time there is two of them?

fixtures/Bridge/Doctrine/Entity/DummyWithIdentifier.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+class DummyWithIdentifier
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
```
theofidry: extra blank line

src/Exception/ObjectGeneratorPersisterException.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Fidry\AliceDataFixtures\Exception;
+
+class ObjectGeneratorPersisterException extends \Exception
```
theofidry: I would make this class `final` and add a `@private` tag to its phpdoc block as it would be for internal usage only

theofidry: I would also extend `\LogicException` and important the class with use statement like any class. The reasoning for `\LogicException` is that it's not a graceful error one can recover from: it's a config error that requires manual change from the user

src/Exception/ObjectGeneratorPersisterException.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static function entityMissingAssignedIdForField($entity)
+    {
+        return new self("Entity of type " . get_class($entity) . " is missing an assigned ID. " .
```
theofidry: You can use doc blocks here:

```
No ID found for the entity "EntityName". If this entity has an auto ID generator, this may be due to having it disabled because once instance of the entity had an an ID assigned. Either remove this assigned ID to allow the auto ID generator to operate or generate and ID for all the "EntityName" entities.
```

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -88,10 +93,41 @@ public function testCanPersistAnEntity($entity, bool $exact = false)
...
+        $dummy->id = 200;
+        $this->persister->persist($dummy);
+
+        $this->persister->flush();
```
theofidry: I would just add a check that the assigned ID is the expected one

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -88,10 +93,41 @@ public function testCanPersistAnEntity($entity, bool $exact = false)
...
+    }
+
+    /**
+     * @expectedException \Fidry\AliceDataFixtures\Exception\ObjectGeneratorPersisterException
```
theofidry: I would also add a `@expectedExceptionMessage`

src/Exception/ObjectGeneratorPersisterException.php
```diff
@@ -0,0 +1,42 @@
...
+
+use LogicException;
+
+/**
+ * @private
+ */
+final class ObjectGeneratorPersisterException extends LogicException
+{
+    /**
+     * No ID found for the entity "EntityName". If this entity has an auto ID generator,
```
theofidry: no sorry I meant as the error message instead of the one bellow :)

I also forgot, instead of making this an exception, I think we should make it a factory. So we could rename this to `*Factory`, and this `entityMissingAssignedIdForField($entity)` method could become `createForEntityMissingAssignedIdForField($entity): LogicException`.

We don't need a custom exception: nobody should catch this one as it requires a manual change from the user

theofidry:
One last minor change and we're ready to go :)

theofidry:
Thanks very much for the hard work :)

dkarlovi:
Very nice work @Nijusan :+1:

PR #91, opened 2018-04-06
Make ID generator easier to override

umpirsky:
As discussed in https://github.com/theofidry/AliceDataFixtures/pull/62#issuecomment-378942905.

### DIFF
--- a/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
+++ b/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
@@ -23,10 +23,7 @@ use Fidry\AliceDataFixtures\Exception\ObjectGeneratorPersisterExceptionFactory;
 use Fidry\AliceDataFixtures\Persistence\PersisterInterface;
 use Nelmio\Alice\IsAServiceTrait;
 
-/**
- * @final
- */
-/*final*/ class ObjectManagerPersister implements PersisterInterface
+class ObjectManagerPersister implements PersisterInterface
 {
     use IsAServiceTrait;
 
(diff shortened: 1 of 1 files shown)


src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -127,6 +126,12 @@ private function getPersistableClasses(ObjectManager $manager): array
         return $persistableClasses;
     }
 
+    protected function setIdGenerator(ORMClassMetadataInfo $metadata): void
```
theofidry: `configureIdGenerator`? I tend to hate `set*` :P

theofidry: you can remove the `@final` tag and comment as it's no longer true, I would however make the public method `persist` final as it does not have much sense to override it

theofidry:
The build failure looks unrelated but it's not good news that it's failing :(

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
umpirsky: Removed, but did not make `persist()` final. I found a comment in some issue which encourages to override it because of missing entity ID issue.

theofidry: Fair, it's fine then

theofidry:
heads up: I'll try to include this in the next release

umpirsky:
Great, thanks!

theofidry:
Closed by #101

PR #96, opened 2018-08-01
Allow multiple doctrine ORM instance

### LINKED ISSUE #90: Multiple ObjectManagers with doctrine
Hi,
i think i found a conceptional bug with the Service Definition of the object manager persister:
```xml
                 class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                 lazy="true">
            <argument type="service" id="doctrine.orm.entity_manager" />
        </service>
```

Having this service definition a `persist(object)` in the Fidry\AliceDataFixtures\Loader\PersisterLoader::load causes the object to be not persistet to the database if it is not in the entity namespace of the default object manager. This is because doctrine.orm.entity_manager is an alias for the default connection. 

One solution would be to add all existing entity managers and \Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister and determine by the getPersistableClasses method which object manager is used for persisting. The only problem with this solution is that an entity can be used in multiple managers. But this may be one of the more rare cases i guess...

jdeniau:
Fixes #90

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/FidryAliceDataFixturesExtension.php
+++ b/src/Bridge/Symfony/DependencyInjection/FidryAliceDataFixturesExtension.php
@@ -16,14 +16,21 @@ namespace Fidry\AliceDataFixtures\Bridge\Symfony\DependencyInjection;
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Doctrine\Bundle\MongoDBBundle\DoctrineMongoDBBundle;
 use Doctrine\Bundle\PHPCRBundle\DoctrinePHPCRBundle;
+use Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister;
+use Fidry\AliceDataFixtures\Bridge\Doctrine\Purger\Purger;
 use Fidry\AliceDataFixtures\Bridge\Symfony\FidryAliceDataFixturesBundle;
+use Fidry\AliceDataFixtures\Loader\PersisterLoader;
+use Fidry\AliceDataFixtures\Loader\PurgerLoader;
 use Fidry\AliceDataFixtures\ProcessorInterface;
 use LogicException;
 use Nelmio\Alice\Bridge\Symfony\NelmioAliceBundle;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Definition;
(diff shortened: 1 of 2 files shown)


jdeniau:
I don't really know why the tests fails. It does not seems related to this PR

theofidry:
Thanks for the PR!

The tests seem to fail due to a new release of the Doctrine packages which introduces more deprecation notices. No biggie, but that would need to be fixed in a separate PR.

Regarding the PR itself, I'm having a bit of trouble to understand: so the strategy is to "duplicate" the current services per entity manager and then you leave the responsibility to the user to pick the right one?

jdeniau:
@theofidry Yes, I used the same strategy that doctrine uses.

Imagine you have the following doctrine configuration:
```yaml
doctrine:
  orm:
    default_entity_manager: foo
    entity_managers:
      foo: 
        connection: xxx
        # etc.
      bar: 
        connection: xxx
        # etc.
```

Then doctrine bundle will generate two services named: `doctrine.orm.foo_entity_manager` and `doctrine.orm.bar_entity_manager` and an alias named `doctrine.orm.entity_manager` for the service `doctrine.orm.foo_entity_manager`. (and it will also generate 2 services for each cache, entity_listener, etc.).

I applied the same strategy here:
Two services named `fidry_alice_data_fixtures.doctrine.foo_purger_loader` and `fidry_alice_data_fixtures.doctrine.bar_purger_loader` and the old service `fidry_alice_data_fixtures.doctrine.purger_loader` is now an alias on `fidry_alice_data_fixtures.doctrine.foo_purger_loader`

theofidry:
The difference is that when you persist an object with Doctrine, provided you have multiple connections/managers, you want to specify which one as wanting to persist it for _all of them_ is much less likely. In alice however, I think it would be more common the other way around: if you have an entity managed by a manager and another entity manager by another, I think most people would like Alice to figure out which manager should be used for which entity.

So I think a different approach that would work better for most cases would be as @dkarlovi mentioned on Slack:

>- fetching all managers
>- asking if they are managing the class in question
>- if so, call `persist()` and mark them dirty
>- repeat for each fixture loaded
>- flush all dirty managers (edited)

Maybe @dkarlovi can also help you out with the implementation

Zayon:
Any updates on this ? This fix would be appreciated :smile:

jdeniau:
@Zayon I worked on this [here](https://github.com/jdeniau/AliceDataFixtures/tree/jd-feat-multipleDoctrineOrmV2) too, but I did not have time to finish this.

Feel free to compare

theofidry:
Closing as I think @Zayon follow up (#99) is a good take on this

dkarlovi:
Sorry all, I should really queue these tasks better, I should have done a follow-up here, my bad.

PR #98, opened 2018-09-06
Add a gitattribute file to exclude test fixtures

stof:
Including test fixtures in the downloadable archive creates confusion for static analysis tools (IDEs, Scrutinizer, etc...) due to the fake classes in them. For instance, when analyzing my project after the upgrade to Alice 3, Scrutinizer decided that `$this->registry->getManager()` in my code was returning a `\Fidry\AliceDataFixtures\Bridge\Doctrine\ORM\FakeEntityManager` rather than a `Doctrine\ORM\EntityManager`.

These files are not needed in the archive.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+/tests export-ignore
+/fixtures export-ignore
+/vendor-bin export-ignore
+/bin export-ignore
+/phpunit*.xml.dist export-ignore
+/Makefile export-ignore


theofidry:
Thanks; looks like I'll have to look at the CI later

stof:
and actually, I think we could even exclude more things (the `migrations` folder seems related to the fixtures too for instance, and some more root files could get excluded)

stof:
Regarding CI, I think it is because DBAL 2.8 added some deprecation warnings for API changes they want to do for DBAL 3, but the ORM is still using the 2.x deprecated way.

PR #99, opened 2018-09-10
Allow multiple doctrine entity manager instances

Zayon:


src/Bridge/Symfony/Resources/config/doctrine_mongodb_odm.xml
```diff
@@ -71,7 +71,7 @@
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine_mongodb.object_manager_persister"
                  class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                  lazy="true">
-            <argument type="service" id="doctrine_mongodb.odm.document_manager" />
+            <argument type="service" id="doctrine" />
```
theofidry: I suspect this would be a different one no? `doctrine` is for the ORM Registry isn't it?

Same comment for the two occurrences below

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -30,7 +31,8 @@
 {
     use IsAServiceTrait;
 
-    private $objectManager;
+    /** @var array|ObjectManager[]  */
```
theofidry: you can omit `array` here

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -42,9 +44,9 @@
      */
     private $metadata = [];
 
-    public function __construct(ObjectManager $manager)
+    public function __construct(ManagerRegistry $managerRegistry)
```
theofidry: Maybe a more BC friendly way would be to inject directly the managers:

```php
function __construct(ObjectManager ...$managers) {
  $this->managers = $managers;
}
```

Since the service is lazy it shouldn't cause any issue. This may require an extra dependency with the expression language though

theofidry:
Note: I'm trying to fix the tests on master (it should be failures related to deprecation notices, but the tests themselves are fine so you can still see the results locally)

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -53,7 +55,7 @@ public function __construct(ObjectManager $manager)
     public function persist($object)
     {
         if (null === $this->persistableClasses) {
```
jdeniau: You can use `$objectManager = $this->managerRegistry->getManagerForClass($class);` with remove all the complexity of `persistableClass`.

It returns `ObjectManager|null`

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
jdeniau: Having the `ManagerRegistry` allow to call `$this->managerRegistry->getManagerForClass($class);` directly, wich remove a lot of complexity. See below

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -81,14 +83,18 @@ public function persist($object)
                 // Do nothing: not supported.
             }
 
-            try {
-                $this->objectManager->persist($object);
-            } catch (ORMException $exception) {
-                if ($metadata->idGenerator instanceof ORMAssignedGenerator) {
-                    throw ObjectGeneratorPersisterExceptionFactory::createForEntityMissingAssignedIdForField($object);
+            foreach ($this->managers as $manager) {
```
jdeniau: Then, as you have an entity manager, you just need to do `$objectManager->persist($object)`.

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -130,8 +143,11 @@ private function getPersistableClasses(ObjectManager $manager): array
     private function getMetadata(string $class): ClassMetadata
```
jdeniau: And here, you can inject the ObjectManager and get metadata directly with `$objectManager->getMetadata($class)`

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
```diff
@@ -79,7 +79,7 @@
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine.object_manager_persister"
```
jdeniau: You will need to deal with the purger service too

src/Bridge/Symfony/Resources/config/doctrine_phpcr_odm.xml
```diff
@@ -73,7 +73,7 @@
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine_phpcr.object_manager_persister"
                  class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                  lazy="true">
-            <argument type="service" id="doctrine_phpcr.odm.document_manager" />
+            <argument type="service" id="doctrine" />
```
jdeniau: should be `doctrine_phpcr`, same with the purger and purger_factory

src/Bridge/Symfony/Resources/config/doctrine_mongodb_odm.xml
jdeniau: Should be `doctrine_mongodb`, same for the purger and purger_factory

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
Zayon: Yeah @jdeniau is right, having the `ManagerRegistry` is nice to remove some complexity.

What should we do ? `ObjectManager ...$managers` with some complexity but no BC or `ManagerRegistry $managerRegistry` with reduced complexity but BC. Up to you @theofidry .

theofidry: can the registry be instantiated here since we cannot inject it?

Zayon: instantiated ? You mean make a `new ManagerRegistry()` inside the constructor ? it seems hard, see the service declaration : https://github.com/doctrine/DoctrineBundle/blob/master/Resources/config/dbal.xml#L59L65

theofidry: Looking at the implementation: https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php#L156

I don't think iterating over the managers is a bad idea in the end. Maybe you can create a custom `ManagerRegistry` class to be able to cache the results a bit like we are doing for the metadata, which you then can instantiate more easily

Zayon:
I've made an update to the PR, it's still WIP but suggestions are welcomed.

- [x] Test it works with one manager
- [ ] Test it works with multiple managers
- [ ] Change Purgers and PugerFactories
- [ ] Add a phpunit test

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -27,35 +27,45 @@ class ObjectManagerPersister implements PersisterInterface
...
     /**
-     * @var array|null Values are FQCN of persistable objects
+     * @var ClassMetadata[] Entity metadata, FQCN being the key
      */
-    private $persistableClasses;
+    private $metadata = [];
 
     /**
-     * @var ClassMetadata[] Entity metadata, FQCN being the key
+     * @var string[ObjectManager] Keys are FQCN of persistable objects, Values are managers
```
stof: I don't think that's a valid phpdoc type.

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -27,35 +27,45 @@ class ObjectManagerPersister implements PersisterInterface
...
      */
-    private $metadata = [];
+    private $classManagerMap;
 
-    public function __construct(ObjectManager $manager)
+    public function __construct(ObjectManager ...$managers)
```
stof: technically, this allows to pass no manager at all (a variable argument is 0+, not 1+)

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -27,35 +27,45 @@ class ObjectManagerPersister implements PersisterInterface
...
     {
-        $this->objectManager = $manager;
+        $this->managers = $managers;
+
+        foreach ($managers as $manager) {
+            $allMetadata = $manager->getMetadataFactory()->getAllMetadata();
```
stof: loading all metadata of a manager is bad for performance, as this is not an optimized API. The optimized API is the one getting metadata for a given class (as the class name is the cache key).

It would be better to build the map on demand, when being asked for a specific class, rather than trying to build it upfront

src/Bridge/Symfony/Resources/config/doctrine_mongodb_odm.xml
```diff
@@ -71,7 +72,7 @@
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine_mongodb.object_manager_persister"
                  class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                  lazy="true">
-            <argument type="service" id="doctrine_mongodb.odm.document_manager" />
+            <argument>%doctrine_mongodb.odm.document_managers%</argument>
```
stof: that won't work. As you made the constructor variadic, you need to add many arguments.

and btw, this parameters contains the *names* of document managers, not the objects (a parameter cannot contain references to services)

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
```diff
@@ -79,7 +80,7 @@
...
                  class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                  lazy="true">
-            <argument type="service" id="doctrine.orm.entity_manager" />
+            <argument>%doctrine.entity_managers%</argument>
```
stof: same issue here than for the ODM

src/Bridge/Symfony/Resources/config/doctrine_phpcr_odm.xml
```diff
@@ -73,7 +75,7 @@
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine_phpcr.object_manager_persister"
                  class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
                  lazy="true">
-            <argument type="service" id="doctrine_phpcr.odm.document_manager" />
+            <argument>%doctrine_phpcr.odm.document_managers%</argument>
```
stof: and again here

stof:
Btw, the issues I reported are telling me that I don't agree with your checked "Test it works with one manager" (and Travis agrees with me, not with you, as CI is failing)

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
Zayon: Yeah, I don't know if it's possible to set the type of keys

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
stof: Well, if you want to keep a variadic constructor, you will need an argument to add all the relevant references as arguments in your service definition.

Zayon:
@theofidry Any idea on what is the cause of the fail ?

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
theofidry: Thinking about it again, what about have a dedicated class implementing `ObjectManager`?

```php
class ObjectsManager implements ObjectManager {
  function (ManagerRegistry $registry) {
    $this->registry = $registry;
  }

  function getClassMetadata(string $class): Metadata {
    $manager = $this->managerRegistry->getManagerForClass($class)
    $manager->getClassMetadata($class);
  }
}
```

So that manager can take the registry just fine and we can inject this new manager without any BC breaks

theofidry:
No idea, would need to dig into it but cannot at the moment. I would suggest to carry on and I can dig into it after the reviews to see what is causing an issue

jdeniau:
Might be a nice way  of dealing with it.

The interface is quite complex (13 methods), but all are probably not used
if the class is internal

theofidry:
Yes, you could throw a `LogicException` on those at first and we can change it later on whenever we want to

theofidry:
It still needs https://github.com/theofidry/AliceDataFixtures/pull/99/files#r219945408 + making the tests greens, but otherwise yeah we are not far off having it merged

dkarlovi:
Note, my first reason for asking about this was to allow me to load ORM and ODM fixtures at the same time (ODM references ORM).

Looking at `hautelook:fixtures:load`, it explicitly requires an ORM manager which means it's unusable in that regard? Would this mean adding a new command is the only way currently?

dkarlovi:
BTW the ODM documents are already properly constructed by `hautelook:fixtures:load`, they're just not flushed, which is quite unfortunate.

theofidry:
>Would this mean adding a new command is the only way currently?

Yes, and potentially can be switched to be the new default command, i.e. `orm`, `odm` stay specific, but the default and general one loads everything, ORM, ODM and every manager

>BTW the ODM documents are already properly constructed by hautelook:fixtures:load, they're just not flushed, which is quite unfortunate

I think there is an open issue about the ORM not having its manager properly propagated, i.e. the right manager is used for loading but not purging. I'm not aware of an issue for the ODM so if there is one it should be opened

dkarlovi:
Here's a quick hack to illustrate the required points, continuation from our Slack conversation:
https://github.com/hautelook/AliceBundle/pull/440

For the feature to work, the new `DoctrineLoader` will need to have manager registries injected. I can just do the entire thing here, but it does reduce reusability of this code (since it would be doubled).

@theofidry do you have any comments there or here?

/cc @mleczakm

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
dkarlovi: I think `ObjectManager[]` would be the best approach since it allows to push managers from both ORM and ODM, you can treat them the same regardless where they came from.

Neirda24: Hi. I don't know if my comment would help but injecting the registry and getting a fresh manager each time you want to persist will also allow insert errors. Usually as soon as the database throws an error, the manager closes. The only way to re-open the connection is to get a fresh manager from the registry. Keeping in mind this I would suggest to inject the registry instead of all the managers.

jdeniau:
Hi,

This should be fixed in #113

Le lun. 17 août 2020 à 16:30, Adrien Roches <notifications@github.com> a
écrit :

PR #101, opened 2018-09-11
Make the ID generator configuration easier to override

theofidry:
Follow up of #91

/cc @umpirsky


theofidry:
Thank you @umpirsky :)

PR #102, opened 2018-09-11
Exclude more root files from the archive

stof:
Follow-up of #98 to remove more things.

Only things kept are:

- the `src` folder (of course necessary)
- the `LICENSE` file (mandatory)
- the `README.md` file (that could be discussed)
- the `composer.json` file (composer does not actually care about that one, but some third-party tools are relying on it to identify third-party packages instead of relying on the composer metadata about installed packages)

I also cleaned the gitignore file from outdated unignored files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,12 @@
 /tests export-ignore
 /fixtures export-ignore
+/migrations export-ignore
 /vendor-bin export-ignore
 /bin export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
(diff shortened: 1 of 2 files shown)


stof:
btw, it would be great to tag a release to benefit from these exclusion of fixtures

theofidry:
Thanks!

I'll tag a release soon depending of the status of #99, maybe you can have a look at it as well :)

stof:
@theofidry I did it, and I suggest releasing without it, as the current state is not ready.

PR #107, opened 2018-10-17
Managers WIP

dkarlovi:
Required for https://github.com/hautelook/AliceBundle/pull/440


dkarlovi:
@theofidry maybe something like this instead of #99?

PR #108, opened 2018-10-24
Feature/multiple entity managers

theofidry:
Comes as a replacement to #99 and laying some groundwork for #107.

@dkarlovi I'm thinking of using the exact same strategy for registries. Right now we have persister/purger for one manager. This PR adds a tuple for a manager registry and then #107 could use a similar strategy to work with multiple registries. WDYT?

- [ ] Tests
- [ ] Wiring the services

### DIFF
--- a/src/Bridge/Doctrine/Purger/ManagerRegistryPurger.php
+++ b/src/Bridge/Doctrine/Purger/ManagerRegistryPurger.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Fidry\AliceDataFixtures package.
+ *
+ * (c) Théo FIDRY <theo.fidry@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Fidry\AliceDataFixtures\Bridge\Doctrine\Persister;
+
+use function array_map;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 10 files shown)


dkarlovi:
@theofidry this looks good to me. :+1: I started reading the commits one by one and got confused where you undid what the first commit did, you should probably squash the branch since there's no original intent anymore.

theofidry:
Indeed, I initially thought I could re-use some of the code but turned out not so much :)

src/Bridge/Doctrine/Purger/ManagerRegistryPurger.php
```diff
@@ -0,0 +1,74 @@
...
+    public function __construct(ManagerRegistry $registry, PurgeMode $purgeMode = null)
+    {
+        $this->registry = $registry;
+
+        $this->purgers = array_map(
+            function (ObjectManager $manager) use ($purgeMode): PurgerInterface {
+                return new ObjectManagerPurger($manager, $purgeMode);
+            },
+            $registry->getManagers()
+        );
```
kgilden: This part here makes the assumption that data from all managers (and hence databases) should be purged every time.

IMHO this seems counter intuitive when running `bin/console hautelook:fixtures:load --manager=foo`. I certainly wouldn't expect the database of manager `bar` be also purged.

Looking at the problem from a wider perspective, perhaps it would be more sensible to remove the `--manager` option and instead create a purger for only those managers which contain at least a single entity with fixtures defined for it?

theofidry: If that could be done I guess it would be less potential side-effects indeed...

PR #110, opened 2018-11-14
Issue with automatically disabling doctrine id generator

Toflar:
So here's a failing test case for the problem described in https://github.com/theofidry/AliceDataFixtures/pull/68.

I'll try it again in other words though:

Disabling the id generator automatically when loading fixtures is really super handy. It's always needed and really comfortable. I can just simply specify something like this and my `id` is not autogenerated even though I have an autogenerator registered for production:

```yaml
App\Entity\Group:
    group_1:
        id: 44b178b5-ea58-44bf-857f-eb8ce4df1280
```

Now here's where it gets a bit tricky: If you have related entities with `cascade: persist` (which is very common I guess), disabling the id generator happens too late for some of the entities.
Imagine this:

```yaml
App\Entity\Group:
    group_1:
        id: 44b178b5-ea58-44bf-857f-eb8ce4df1280

App\Entity\User:
    user_1:
        id: b3c52f50-992f-44fd-bc71-95acc1eef687
        group: "@group_1"
```

The way the fixture loader works is it persists every object one after the other. So it will persist `group_1` first and only then `user_1`. Now if you do not have `cascade: persist` configured, it works just fine. However, if you configured `cascade: persist` the `id` for `user_1` is generated during `->persist($group1)` already.

So handling this in the `ObjectManagerPersister` does not work. Disabling the UUID generation has to be done **before any** of the objects is being persisted.
So either we need a special loader (which I don't really like because loading is actually not affected) or we have to have a new interface a `Processor` can implement that is called **before** the persister is asked to persist **any** object.

### DIFF
--- a/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
+++ b/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
@@ -22,6 +22,7 @@ use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\DummyEmbeddable;
 use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\DummySubClass;
 use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\DummyWithEmbeddable;
 use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\DummyWithIdentifier;
+use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\DummyWithRelatedCascadePersist;
 use Fidry\AliceDataFixtures\Bridge\Doctrine\Entity\MappedSuperclassDummy;
 use Fidry\AliceDataFixtures\Exception\ObjectGeneratorPersisterException;
 use Fidry\AliceDataFixtures\Persistence\PersisterInterface;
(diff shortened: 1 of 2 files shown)


tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -183,7 +204,8 @@ public function provideEntities()
                 $dummy->id = 300;
 
                 return $dummy;
-            })()
+            })(),
+            true
```
Toflar: I have re-added here because otherwise that test does not make sense there. Not sure why it was removed because it was present in the original PR: https://github.com/theofidry/AliceDataFixtures/pull/68

Toflar:
BTW: For people encountering the same issue. You can temporarily fix it by changing the order of your fixture objects. In my example it would work if I defined `user_1` before `group_1`.

theofidry:
@Toflar is this issue still present with #135?

Toflar:
Sorry Théo but this was so long ago, I can't remember and I must have long fixed it otherwise. I don't even remember which project that was :D

videni:
@@Toflar， this issue still exists

theofidry:
Closed by #198

PR #112, opened 2018-12-05
Fixed root_node deprecation with Symfony 4.2

jmsche:
Hi, this PR fixes a small deprecation introduced in Symfony 4.2; the fix includes a BC layer for older versions.


theofidry:
Since it's not the only commit for the next release and I want to check some stuff before the next release, it will have to wait a bit. This month is a busy month but alice is the first thing I'll check when I'm back to it

lunetics:
@theofidry Is there any way to support you, since 5.0 is out now?

theofidry:
Yes: updating the CI and making it green & maybe adjusting the dependencies accordingly (we can drop some old ones). Then I can tag it

PR #113, opened 2018-12-27
Allow multiple entity manager

jdeniau:
See explanations in code


src/Bridge/Doctrine/Persister/ManagerRegistryPersister.php
```diff
@@ -42,12 +42,12 @@ public function __construct(ManagerRegistry $registry)
         $managers = $registry->getManagers();
 
         foreach ($managers as $manager) {
-            $this->persisters[get_class($manager)] = new ObjectManagerPersister($manager);
```
jdeniau: This allows having multiple instance of `Doctrine\ORM\EntityManager`

src/Bridge/Doctrine/Persister/ManagerRegistryPersister.php
```diff
@@ -80,6 +80,6 @@ private function getPersisterForClass(string $class): PersisterInterface
             );
         }
 
-        return $this->persisters[get_class($manager)];
```
jdeniau: (this matches previous comment)

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -87,7 +87,7 @@ public function persist($object)
                 throw $exception;
             }
 
-            if (null !== $generator && false === $generator->isPostInsertGenerator()) {
```
jdeniau: Found this bug when using a "cascade persist" relationship :

An item (lets say it is a `Category`) persisted in database with an id, the code lines 67 to 72 will remove the generator from the metadata.
As it is a `Doctrine\ORM\Id\IdentityGenerator` (isPostInsertGenerator == true), the generator will not be reset.
After that, you persist an `Article` with a new `Category`, the Category metadata will still be of type `GENERATOR_TYPE_NONE`, and Doctrine will throw an exception because id generation strategy is wrongly set to None.

I just commented the `isPostInsertGenerator` to open discussion because I do not really see the point of this test.
For me, if the generator has been changed, it is needed to reset it to its original value, don't  you think ?

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
```diff
@@ -78,7 +78,7 @@
                  public="true" />
 
         <service id="fidry_alice_data_fixtures.persistence.persister.doctrine.object_manager_persister"
-                 class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister"
+                 class="Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ManagerRegistryPersister"
```
jdeniau: `ObjectManagerPersister` take an EntityManager as parameter, but we give a Registy

src/Bridge/Doctrine/Persister/ManagerRegistryPersister.php
```diff
@@ -42,12 +42,12 @@ public function __construct(ManagerRegistry $registry)
         $managers = $registry->getManagers();
 
         foreach ($managers as $manager) {
-            $this->persisters[get_class($manager)] = new ObjectManagerPersister($manager);
+            $this->persisters[spl_object_hash($manager)] = new ObjectManagerPersister($manager);
```
jdeniau: This allows having multiple instance of `Doctrine\ORM\EntityManager`

src/Bridge/Doctrine/Persister/ManagerRegistryPersister.php
```diff
@@ -80,6 +80,6 @@ private function getPersisterForClass(string $class): PersisterInterface
             );
         }
 
-        return $this->persisters[get_class($manager)];
+        return $this->persisters[spl_object_hash($manager)];
```
jdeniau: (this matches previous comment)

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -87,7 +87,7 @@ public function persist($object)
...
+            if (null !== $generator/* && false === $generator->isPostInsertGenerator()*/) {
```
jdeniau: Found this bug when using a "cascade persist" relationship :

An item (lets say it is a `Category`) persisted in database with an id, the code lines 67 to 72 will remove the generator from the metadata.
As it is a `Doctrine\ORM\Id\IdentityGenerator` (isPostInsertGenerator == true), the generator will not be reset.
After that, you persist an `Article` with a new `Category`, the Category metadata will still be of type `GENERATOR_TYPE_NONE`, and Doctrine will throw an exception because id generation strategy is wrongly set to None.

I just commented the `isPostInsertGenerator` to open discussion because I do not really see the point of this test.
For me, if the generator has been changed, it is needed to reset it to its original value, don't  you think ?

theofidry:
@jdeniau sorry this PR went a bit unnoticed. Would it be possible to update it with the latest changes?

src/Bridge/Doctrine/Persister/ManagerRegistryPersister.php
theofidry: 👍

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
theofidry: Thanks to #135 the generators state have been restored and I think the `isPostInsertGenerator()` has been removed since then

src/Bridge/Symfony/Resources/config/doctrine_orm.xml
theofidry: it's curious that this works...

jdeniau:
@theofidry My branch was originally based upon [your `feature/multiple-entity-managers` branch](https://github.com/theofidry/AliceDataFixtures/compare/feature/multiple-entity-managers).

I rebased this PR on `master`, we now have our mixed work on multiple entity manager mixed.

I had a conflict on the `Purger` class, I picked the changes made [here](https://github.com/theofidry/AliceDataFixtures/commit/a2c2e67043d1868b75b45ab73248ea9e0aa26300#diff-31be3fb4f0b30a6299d21324ed917e07)

jdeniau:
This PR does now superseed #99

theofidry:
👌  Looks like the CI is failing.

I suppose this should be part of a major release? I'm afraid it would introduce some BC breaks...

jdeniau:
> 👌 Looks like the CI is failing.
> 
> I suppose this should be part of a major release? I'm afraid it is introducing some BC breaks...

Hi, I fixed the CI, the bugs came from the `Purger` class that now extends `ObjectManagerPurger` and thus the private properties were not enabled anymore.

The only failing test is the same failing test as master.

Theoretically, there should be no BC break, as the XML files are updated.
But I am not 100% either, and an internal refactoring might be a valid argument to create a major release, so it looks good to me.

jdeniau:
For the record if you push a major release, you might want to drop the deprecated classes (`Purger` for example).
There might be some cleaning to do after thi PR is merged in that case

jdeniau:
@theofidry I took time to re-work on that PR again. 
The CI is failing but for the same reason as https://github.com/theofidry/AliceDataFixtures/pull/168 . Probably because the lock file is not set and some dependency did change since last commit on master ? 🤔 

I will try to test intensively in the next days with doctrine/orm, and I come back to you if everything is OK.

COil:
> @theofidry I took time to re-work on that PR again.
> The CI is failing but for the same reason as #168 . Probably because the lock file is not set and some dependency did change since last commit on master ? 🤔
> I will try to test intensively in the next days with doctrine/orm, and I come back to you if everything is OK.

Hi jdeniau, did you have time to make progress on this PR? I am discovering this right now, and I'll be interested in this feature too. See you.

jdeniau:
Hi @COil , 

I ran into a lot of new issues with this PR, probably due to https://github.com/theofidry/AliceDataFixtures/pull/155 and the entity id management.

As I was upgrading the whole package of behat / alice and the all ecosystem ([Symfony2Extension](https://github.com/Behat/Symfony2Extension) to [SymfonyExtension](https://github.com/FriendsOfBehat/SymfonyExtension) for the main upgrade), I did not manage to make this work, and won't take anymore time on it.

The solution I came by is the following though:

```php
<?php

declare(strict_types=1);

namespace Mapado\TicketingBundle\Tests\Context;

use App\Entity\SomeEntity;
use Behat\Behat\Context\Context;
use Behat\Gherkin\Node\TableNode;
use Doctrine\ORM\Mapping\ClassMetadataInfo;
use Doctrine\Persistence\ManagerRegistry;
use Nelmio\Alice\FileLoaderInterface;
use Symfony\Component\Yaml\Yaml;

class FixtureLoaderContext implements Context
{
    // we were probably messing up with ids. I managed to remove nearly all ids from fixtures, 
    // but for some files, it was just too complex.
    private const CLASS_WITH_FORCED_IDS = [
        SomeEntity::class,
        // …
    ];

    /**
     * @var FileLoaderInterface
     */
    private $aliceFileLoader;

    /**
     * @var string
     */
    private $basePath;

    /** @var ManagerRegistry */
    private $doctrine;

    public function __construct(
        FileLoaderInterface $aliceFileLoader,
        ManagerRegistry $doctrine,
        string $basePath
    ) {
        $this->aliceFileLoader = $aliceFileLoader;
        $this->doctrine = $doctrine;
        $this->basePath = $basePath;
    }

    /**
     * @SuppressWarnings(PHPMD.UnusedFormalParameter("persister"))
     *
     * @Given the following fixtures are loaded:
     * @Given the following fixtures files are loaded:
     * @Given the following fixtures files are loaded with the persister :persister:
     *
     * @param TableNode<string> $fixtures Path to the fixtures
     */
    public function thereAreSeveralFixtures(TableNode $fixtures, ?string $persister = null): void
    {
        $files = [];
        foreach ($fixtures->getRows() as $row) {
            $files[] = $row[0];
        }

        $this->loadFiles($files);
    }

    /**
     * @SuppressWarnings(PHPMD.UnusedFormalParameter("persister"))
     *
     * @Given the fixtures file :file is loaded with the persister :persister
     */
    public function theFixturesFileIsLoadedWithThePersister(string $file, ?string $persister = null): void
    {
        $this->loadFiles([$file]);
    }

    /**
     * TODO keep this method in case we want to load all file simultaneously (or see following TODO)
     *
     * @param array<string> $files
     */
    private function loadFiles(array $files): void
    {
        $fixturesFiles = [];
        $objects = [];

        foreach ($files as $file) {
            $fixturesFile = $this->basePath . $file;

            $fixturesFiles[] = $fixturesFile;

            // TODO this is probably useless as "do load file" does flush on each file now
            // but it was a list of files that I wanted to be flushed in DB BEFORE other files do load, as we have some db id fetching
            if (in_array($file, ['10.some_file.yml'])) {
                $objects = $this->doLoadFiles($fixturesFiles, $objects);
                $fixturesFiles = [];
            }
        }

        if ($fixturesFiles) {
            $this->doLoadFiles($fixturesFiles, $objects);
        }
    }

    /**
     * @param array<string> $files
     * @param array<object> $objects
     */
    private function doLoadFiles(array $files, array $objects)
    {
        // force offer rule id to be set
        foreach (self::CLASS_WITH_FORCED_IDS as $class) {
            $entityManager = $this->doctrine->getManagerForClass($class);
            if (!$entityManager) {
                throw new \RuntimeException('Unable to get entity manager for class ' . $class);
            }
            $metadata = $entityManager->getClassMetadata($class);
            if (!$metadata instanceof ClassMetadataInfo) {
                throw new \RuntimeException('Class with forced id metata must implement ClassMetadataInfo. This should not happen.');
            }

            // I force the `GENERATOR_TYPE_NONE` here to keep the old alice data fixture compatibility
            // You may want to use something different here ¯\_(ツ)_/¯
            $metadata->setIdGeneratorType(\Doctrine\ORM\Mapping\ClassMetadata::GENERATOR_TYPE_NONE);
        }

        $this->preventObjectNameConflict($files);

        $loader = $this->aliceFileLoader;

        $parameters = [];
        $objectList = $objects;
        foreach ($files as $file) {
            $objectSet = $loader->loadFile($file, $parameters, $objectList);

            $parameters = $objectSet->getParameters();
            $objectList = $objectSet->getObjects();

            foreach ($objectList as $object) {
                /** @var class-string */
                $objectClass = get_class($object);

                // if object with forced id, check that the id is set
                if (in_array($objectClass, self::CLASS_WITH_FORCED_IDS) && !$object->getId()) {
                    throw new \InvalidArgumentException($objectClass . ' ids must be set in fixtures in file ' . $file);
                }

                $entityManager = $this->doctrine->getManagerForClass($objectClass);

                if (!$entityManager) {
                    throw new \RuntimeException('Unable to find manager for class ' . $objectClass);
                }

                $entityManager->persist($object);
            }

            $entityManager->flush();
        }

        return $objectList;
    }

    /**
     * This function makes sur that there is no key present in several files loaded in the same time both files will be loaded in the same array
     * 
     * @SuppressWarnings(PHPMD.UnusedLocalVariable("localObject"))
     *
     * @param array<string> $files
     */
    private function preventObjectNameConflict(array $files): void
    {
        $objectList = [];
        foreach ($files as $file) {
            $parsedYaml = Yaml::parseFile($file);

            if ($parsedYaml) {
                foreach ($parsedYaml as $localObjectList) {
                    foreach ($localObjectList as $localKey => $localObject) {
                        if (array_key_exists($localKey, $objectList)) {
                            throw new \InvalidArgumentException(
                            sprintf(
                                'key %s of file %s already exist in previously loaded objects',
                                $localKey,
                                $file
                            )
                        );
                        }
                    }
                    $objectList += $localObjectList;
                }
            }
        }
    }
}
```

If you want to continue working on this PR, feel free to do so.

COil:
Hello @jdeniau , thanks for the answer and the code. I came with another solution where I decorate the `ObjectManagerPersister` service. In the `persist()` function I also use `getManagerForClass` the get the good entity manager to use, then I force the `ObjectManagerPersister.objectManager` property with reflection (the `persistableClasses` property must also be reinitialized when the entity manager is changed).

PR #121, opened 2019-07-31
Disable FK checks also on DELETE

tifabien:
Actually I'm facing the same issue with the DELETE purge mode. I got the following error when trying to refresh my database.

> In AbstractMySQLDriver.php line 49:
>                                                                                                                               
>   An exception occurred while executing 'DELETE FROM portal':                                                                 
>                                                                                                                               
>   SQLSTATE[23000]: Integrity constraint violation: 1217 Cannot delete or update a parent row: a foreign key constraint fails

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -20,7 +20,7 @@
          verbose="true">
 
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak_vendors" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="0" />
     </php>
 
     <testsuites>
--- a/phpunit_doctrine.xml.dist
+++ b/phpunit_doctrine.xml.dist
@@ -20,7 +20,7 @@
          verbose="true">
 
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak_vendors" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="0" />
 
         <env name="DB_DRIVER" value="pdo_mysql" />
         <env name="DB_USER" value="root" />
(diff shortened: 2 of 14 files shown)


theofidry:
Thanks for the PR!

- Would it be possible to make sure the CI is green before that? It seems to be failing due to a deprecation error message which could be fixed beforehand
- Any chance to get a test for it? IIRC there is one already partially implemented for it

tifabien:
> * Would it be possible to make sure the CI is green before that? It seems to be failing due to a deprecation error message which could be fixed beforehand

I can try indeed to get the CI to pass. I got one small question because it seems the "weak_vendors" has been deprecated since SF 4.3. How can I do to keep this "weak_vendors" when testing with SF < 4.3 and testing with the new "max[self]=0" when SF 4.3+ is used?

theofidry:
Maybe we can allow a certain number of deprecations instead?

tifabien:
Do you prefer I remove the `SYMFONY_DEPRECATIONS_HELPER` env in all phpunit.xml.dist file (as 0 is the default value) or I change the value to `0` instead of `weak_vendors`

theofidry:
Setting it to `0` looks more explicit to me 👍

tifabien:
@theofidry I think I'm done now :) CI is now green. I updated values for all `SYMFONY_DEPRECATIONS_HELPER` env + I updated dependency of proxyManager (the last version supports PHP 7.4)

theofidry:
Thanks a lot, sorry somehow forgot about that PR :(

tifabien:
@theofidry Do you think it's possible to include this PR in a new release or do you have something on the roadmap in the repo?

theofidry:
There's 242497e I'm not entirely sure of the effects. But ultimately I kinda wanted to have https://github.com/theofidry/AliceDataFixtures/pull/108 in tbh

PR #126, opened 2019-11-24
Fix tests

alexislefebvre:
Field "embeddable" in class "Fidry\AliceDataFixtures\Bridge\Doctrine\MongoDocument\DummyWithEmbeddable"
relies on same-namespace resolution for the target document.
This is deprecated and will be dropped in doctrine/mongodb-odm 2.0. Please use a FQCN instead.


alexislefebvre:
It only fix a test from `make test_doctrine_odm_bridge`.

PR #127, opened 2019-11-28
Fix test_doctrine_odm_bridge test

alexislefebvre:
It only fix a test from `make test_doctrine_odm_bridge`.

I didn't managed to fix the other tests. I think that this can still be merged as is.


theofidry:
Thanks @alexislefebvre;

Note that if you think that would make it easier, the databases could be migrated to docker instead

alexislefebvre:
I agree, today it requires installing dependencies like `mongo` in our environment. It would be much more convenient to launch containers for MySQL, MongoDB, etc. for tests.

Issue #123, opened 2019-09-28
TreeBuilder::root() is deprecated since Symfony 4.3


jianzhong:
After installation of this package, when I run phpunit, I got this warning message:
The "Symfony\Component\Config\Definition\Builder\TreeBuilder::root()" method called for the "fidry_alice_data_fixtures" configuration is deprecated since Symfony 4.3, pass the root name to the constructor instead.

MLDMoritz:
A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0!

gharlan:
already fixed in https://github.com/theofidry/AliceDataFixtures/pull/112

mpiot:
Hi, is there a date scheduled for tagging a new version ? Symfony 5.0 is coming (november) and this patch was merged more than 10 months ago :/

theofidry:
Unless someone is willing to work on #108 I'm fine with tagging a new version but IMO the CI should be updated to be green on both PHP 7.3 and 7.4 before that.

ro0NL:
friendly ping :)

snebes:
Another friendly ping, especially with Symfony 5.0 released today

theofidry:
As stated [here](https://github.com/theofidry/AliceDataFixtures/issues/123#issuecomment-545955388), I'm fine with tagging it but IMO the CI should be fixed first

alexislefebvre:
> Unless someone is willing to work on #108 I'm fine with tagging a new version but IMO the CI should be updated to be green on both PHP 7.3 and 7.4 before that.

The CI is green after the merge of #128. Is it possible to publish a new release that fix the deprecation?

theofidry:
Done: https://github.com/theofidry/AliceDataFixtures/releases/tag/v1.1.2

PR #135, opened 2020-05-06
Avoid permanent change of the IdGenerator entity metadata

### LINKED ISSUE #105: Entity of type [EntityClass] is missing an assigned ID for field 'id'
After a fresh upgrade to v1.1, I have this error on a functional test:

```
PHPUnit 7.2.7 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 5.74 seconds, Memory: 149.00MB

There was 1 failure:

1) Tests\AppBundle\Command\DomainNameExpireCheckCommandTest::test
Failed asserting that '\n
In ORMException.php line 87:\n
                                                                                                                                                                                                                                                                               \n
  Entity of type AppBundle\Entity\Ticket is missing an assigned ID for field  'id'. The identifier generation strategy for this entity requires the ID field to be populated before EntityManager#persist() is called. If you want automatically generated identifiers instea  \n
  d you need to adjust the metadata mapping accordingly.                                                                                                                                                                                                                       \n
                                                                                                                                                                                                                                                                               \n
\n
nexy:domain-name:expire-check [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command>\n
\n
' contains "Expire ticket created for expiring.com.".

/code/tests/AppBundle/Command/DomainNameExpireCheckCommandTest.php:98

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

Git bisect revealed this commit: 7053c75de4d5a91c5bd7db43918a907706c7c990 coming from #89.

The Ticket entity id field:

```php
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Table
 * @ORM\Entity(repositoryClass="AppBundle\Repository\TicketRepository")
 * @ORM\EntityListeners({"AppBundle\Listener\Entity\TicketEntityListener"})
 */
class Ticket
{
    /**
     * @var int
     *
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;
```

The test file:

```php
use AppBundle\Entity\DomainName;
use AppBundle\Entity\Purchase;
use Doctrine\ORM\EntityManagerInterface;
use PowerDNSBundle\Doctrine\ORM\PowerDNSEntityManager;
use PowerDNSBundle\Entity\PowerDNSDomain;
use Tests\AppBundle\Controller\WebTestCase;

final class DomainNameAutoRenewCommandTest extends WebTestCase
{
    /**
     * @var EntityManagerInterface
     */
    private $entityManager;

    /**
     * @var PowerDNSEntityManager
     */
    private $powerDnsManager;

    protected function setUp(): void
    {
        parent::setUp();

        $this->entityManager = $this->getContainer()->get('doctrine.orm.default_entity_manager');
        $this->powerDnsManager = $this->getContainer()->get(PowerDNSEntityManager::class);

        $purchase = (new Purchase())
            ->setVersion(1)
            ->setPrice(15.4)
            ->setNextInvoiceDate(new \DateTime('tomorrow'))
            ->setProduct($this->getReference('product_domain_name'))
            ->setCustomer($this->getReference('user_default'))
            ->setAccounting($this->getReference('accounting_fr'))
        ;

        $domainName = (new DomainName())
            ->setFqdn('renewal.com')
            ->setAutoRenew(true)
            ->setStatus(DomainName::STATE_CREATED)
            ->setExpirationDate(new \DateTime('+1 week'))
            ->setOwnerContact('TEST45-GANDI')
            ->setAdministratorContact('TEST45-GANDI')
            ->setBillingContact('TEST45-GANDI')
            ->setTechnicianContact('TEST45-GANDI')
            ->setPurchase($purchase)
        ;

        $this->entityManager->persist($purchase);
        $this->entityManager->persist($domainName);
        $this->entityManager->flush();

        $powerDnsDomain = (new PowerDNSDomain())
            ->setName($domainName->getFqdn())
            ->setOwner($this->getReference('user_default'))
            ->setDomainName($domainName)
            ->setDomainNameId($domainName->getId())
        ;

        $this->powerDnsManager->persist($powerDnsDomain);
        $this->powerDnsManager->flush();
    }

    public function test(): void
    {
        $output = $this->runCommand('nexy:domain-name:auto-renew');

        $this->assertContains('Renewing renewal.com.', $output);

        $output = $this->runCommand('nexy:domain-name:auto-renew');

        $this->assertNotContains('Renewing renewal.com.', $output);
    }

    /**
     * {@inheritdoc}
     */
    protected function getFixturesFiles(): array
    {
        return \array_merge(parent::getFixturesFiles(), [
            __DIR__.'/../../fixtures/orm/product.yml',
            __DIR__.'/../../fixtures/orm/accounting.yml',
        ]);
    }
}
```

Tell me if you need anything more to investigate.

soullivaneuh:
fixes #105

In any case the generator is changed on the metadata, it must be restored afterward.

Otherwise, manual entity persistence after fixture loading will cause troubles.

Tests was added during the feature addition: https://github.com/theofidry/AliceDataFixtures/commit/7053c75de4d5a91c5bd7db43918a907706c7c990#diff-292d5068b7f30308701962493b4cfd57

So if Travis is green, that will prove the condition is not necessary.

Otherwise, we have to find an another way to deal with it, because leaving the metadata at this state is not acceptable.

### DIFF
--- a/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
+++ b/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
@@ -35,9 +35,9 @@ class ObjectManagerPersister implements PersisterInterface
     private $persistableClasses;
 
     /**
-     * @var ClassMetadata[] Entity metadata, FQCN being the key
+     * @var ClassMetadata[] Entity metadata to restore after flush, FQCN being the key.
      */
-    private $metadata = [];
+    private $metadataToRestore = [];
 
     public function __construct(ObjectManager $manager)
     {
@@ -56,19 +56,13 @@ class ObjectManagerPersister implements PersisterInterface
         $class = get_class($object);
 
         if (isset($this->persistableClasses[$class])) {
-            $metadata = $this->getMetadata($class);
-
-            $generator = null;
-            $generatorType = null;
+            $metadata = $this->objectManager->getClassMetadata($class);
 
             // Check if the ID is explicitly set by the user. To avoid the ID to be overridden by the ID generator
             // registered, we disable it for that specific object.
             if ($metadata instanceof ORMClassMetadataInfo) {
                 if ($metadata->usesIdGenerator() && false === empty($metadata->getIdentifierValues($object))) {
-                    $generator = $metadata->idGenerator;
-                    $generatorType = $metadata->generatorType;
-
-                    $this->configureIdGenerator($metadata);
+                    $metadata = $this->configureIdGenerator($metadata);
                 }
             } elseif ($metadata instanceof ODMClassMetadataInfo) {
                 // Do nothing: currently not supported as Doctrine ODM does not have an equivalent of the ORM
@@ -86,12 +80,6 @@ class ObjectManagerPersister implements PersisterInterface
 
                 throw $exception;
             }
-
-            if (null !== $generator && false === $generator->isPostInsertGenerator()) {
-                // Restore the generator if has been temporary unset
-                $metadata->setIdGeneratorType($generatorType);
-                $metadata->setIdGenerator($generator);
-            }
         }
     }
 
@@ -101,6 +89,11 @@ class ObjectManagerPersister implements PersisterInterface
     public function flush()
     {
         $this->objectManager->flush();
+
+        foreach ($this->metadataToRestore as $metadata) {
+            $this->objectManager->getMetadataFactory()->setMetadataFor($metadata->getName(), $metadata);
+        }
+        $this->metadataToRestore = [];
     }
 
     /**
@@ -123,19 +116,23 @@ class ObjectManagerPersister implements PersisterInterface
         return $persistableClasses;
     }
 
-    protected function configureIdGenerator(ORMClassMetadataInfo $metadata): void
+    private function saveMetadataToRestore(ClassMetadata $metadata): void
     {
-        $metadata->setIdGeneratorType(ORMClassMetadataInfo::GENERATOR_TYPE_NONE);
-        $metadata->setIdGenerator(new ORMAssignedGenerator());
+        if (!isset($this->metadataToRestore[$metadata->getName()])) {
+            $this->metadataToRestore[$metadata->getName()] = $metadata;
+        }
     }
 
-    private function getMetadata(string $class): ClassMetadata
+    private function configureIdGenerator(ORMClassMetadataInfo $metadata): ORMClassMetadataInfo
     {
-        if (false === array_key_exists($class, $this->metadata)) {
-            $classMetadata = $this->objectManager->getClassMetadata($class);
-            $this->metadata[$class] = $classMetadata;
-        }
+        $this->saveMetadataToRestore($metadata);
 
-        return $this->metadata[$class];
+        $newMetadata = clone $metadata;
+        $newMetadata->setIdGeneratorType(ORMClassMetadataInfo::GENERATOR_TYPE_NONE);
+        $newMetadata->setIdGenerator(new ORMAssignedGenerator());
+
+        $this->objectManager->getMetadataFactory()->setMetadataFor($metadata->getName(), $newMetadata);
+
+        return $newMetadata;
     }
 }
--- a/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
+++ b/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
@@ -107,7 +107,17 @@ class ObjectManagerPersisterTest extends TestCase
         $dummy->id = 200;
         $this->persister->persist($dummy);
 
+        $this->assertFalse(
+            $this->entityManager->getClassMetadata(DummyWithIdentifier::class)->usesIdGenerator(),
+            'ID generator should be changed.'
+        );
+
         $this->persister->flush();
+
+        $this->assertTrue(
+            $this->entityManager->getClassMetadata(DummyWithIdentifier::class)->usesIdGenerator(),
+            'ID generator should be restored after flush.'
+        );
 
         $entity = $this->entityManager->getRepository(DummyWithIdentifier::class)->find(200);
         $this->assertInstanceOf(DummyWithIdentifier::class, $entity);


soullivaneuh:
> So if Travis is green

Well, looks like Travis does not give a single f*ck of my PR.

theofidry:
>Otherwise, we have to find an another way to deal with it, because leaving the metadata at this state is not acceptable.

I'm not 100% sure to understand how the generator metadata works to be honest: if its state matters before the persist or if it does until the flush. If the later, we can restore it in the `flush()` instead

For the tests I don't know why Travis is not being triggered... Did you try `make test` locally?

soullivaneuh:
> If the later, we can restore it in the flush() instead

We don't have entity access on flush event. Are you suggesting to store them on the class to reset them later? That's a possibility, but are we sure we need this? What the condition I removed does exactly?

> For the tests I don't know why Travis is not being triggered... Did you try make test locally?

The `make test` require a lot of system dependencies to install, it's quite hard to make it running locally.

Do you have access to the Travis admin panel of this project? Maybe the integration is broken.

theofidry:
>Do you have access to the Travis admin panel of this project? Maybe the integration is broken.

There's the webhook and it seems to be working fine. That said I saw Travis behaving weirdly those last few weeks so it might have some trouble right now...

>We don't have entity access on flush event. Are you suggesting to store them on the class to reset them later? That's a possibility, but are we sure we need this? What the condition I removed does exactly?

I'm not sure what the condition you removed does exactly. But yeah we could store the necessary information to restore the metadata later on. I don't have a strong opinion about which solution should be used: I think it mostly depends on whether or not the metadata is used between the persist and the flush

soullivaneuh:
I propose to let you find why Travis is not working anymore first.

Then if test are green, just use this solution. If this lead to another errors we will be able to correctly identify them and add comments with the condition, and we will use the flush method.

What do you think?

theofidry:
Ok let's do that

soullivaneuh:
Travis job looks launched: https://travis-ci.org/github/theofidry/AliceDataFixtures/builds/683853102

Maybe a configuration issue with the GitHub repository? Try to re-enable the project on Travis, it may fix the problem.

theofidry:
Double checked and re-enabled

soullivaneuh:
Tests are failing. Not really sure what @Nijusan wanted to do with this condition (maybe he will respond!), but I maybe find a cleaner solution to restore all at flush.

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -123,19 +116,23 @@ private function getPersistableClasses(ObjectManager $manager): array
...
+    private function saveMetadataToRestore(ClassMetadata $metadata): void
     {
-        $metadata->setIdGeneratorType(ORMClassMetadataInfo::GENERATOR_TYPE_NONE);
-        $metadata->setIdGenerator(new ORMAssignedGenerator());
+        if (!isset($this->metadataToRestore[$metadata->getName()])) {
+            $this->metadataToRestore[$metadata->getName()] = $metadata;
+        }
     }
 
-    private function getMetadata(string $class): ClassMetadata
```
soullivaneuh, line 132: This metadata cache is useless has doctrine already has one: https://github.com/doctrine/orm/blob/9a8e1fb47776fc5ff49f995f3fef93a5ff8469b0/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L239-L241

Removed.

src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
```diff
@@ -123,19 +116,23 @@ private function getPersistableClasses(ObjectManager $manager): array
...
+    private function configureIdGenerator(ORMClassMetadataInfo $metadata): ORMClassMetadataInfo
```
soullivaneuh, line 126: The `configureIdGenerator` has no reason to be `protected` AFAIK, so I make it private and changed its signature.

soullivaneuh:
And voilà! Did a clone of the metadata and replaced it with the new one.

Then, restored and flush.

Also simplify a bit the codebase and added a proof test.

I leave some explanations on the diff. Ready for review. :+1:

soullivaneuh:
Hmm looks like travis jobs are not triggered at all now: https://travis-ci.org/github/theofidry/AliceDataFixtures/pull_requests

And the last job of #136 is failing. So failing on this one may not be related.

I may take a look to the settings with granted access if you want to.

soullivaneuh:
Ran `make test_doctrine_bridge` successfully on my local.

theofidry:
Will take a fresh look tomorrow

Nijusan:
> Tests are failing. Not really sure what @Nijusan wanted to do with this condition (maybe he will respond!), but I maybe found a cleaner solution to restore all at flush.

hey @soullivaneuh , sorry for late response, which condition do u mean? long time ago i worked on this projekt :)

theofidry:
@soullivaneuh could you try to push an empty commit now to see if that works?

theofidry:
@soullivaneuh Travis should be fixed and Symfony 5 support has been merged

Nijusan:
> Tests are failing. Not really sure what @Nijusan wanted to do with this condition (maybe he will respond!), but I maybe found a cleaner solution to restore all at flush.

maybe it's still useful information: i recently updated a project from symfony 4.4 to 5.1 and ran into problems after loading my fixtures. and i remembered why i added this condition: it was not possible to mix fixtures for one entity some with given id-information and some relying on a post insert generator. so it has to be all or nothing. either all fixtures of this entity have  the id set in fixtures or none of them. 

this solution worked fine until the update. now problems occur in a specific phpunit test, where first fixtures of one entity are loaded with given id information, and in the test another new entity is tried to persist and fails then with wrong parameter exception...

theofidry:
@Nijusan I'm not sure to understand the solution you're proposing here: you mean the change to the metadata should be done right before the flush (right now it's done during the first pass) and restored after (like done in this PR)? Or are you suggesting to use another hook?

Nijusan:
@theofidry sorry for the confusion. i just wanted to point out why i implemented this check. given an entity with primary key id, it was not possible to have a fixture file with (in the case of a post insert id generator):

```
Entity\MyEntity:
    my_entity_1:
        id: 3
       name: test

    my_entity_2:
        name: foo
```
 if i restore the metadata in this case before the flush, the given id: 3 from my fixture wont be used. and because of that, we said if we want to manual define the primary key, all given entities have to do so.

so i think the right thing is to restore the metadata after the flush (like done in this PR), still having the problem with all or nothing (all entites have explicit primary kes in fixture file or none of them). but i tried the changes in my project without success, not all fixtures were loaded correctly :( I'll try to take some time to check why, maybe it was just a mistake of mine

theofidry:
Thank you @soullivaneuh

theofidry:
@Nijusan yes it's an all or nothing right now, but this is more a Doctrine limitation than alice unfortunately

yakobe:
Any chance we can get this released? 🙂

theofidry:
done

yakobe:
Thank you!

tristanbes:
I can confirm that this PR breaks previously working fixtures with error:

```
No ID found for the entity "App\Entity\Config". If this entity has an auto ID generator, this may be due to having it disabled because one instance of the entity had an ID assigned. Either remove this assigned ID to allow the auto ID generator to operate or generate and ID for all the "App\Entity\Config" entities
```

For the given fixture:

```yml
App\Entity\Config:
    config_etude_lyon_fidgar:
        name: 'Fid Garage'
        codeService: 'FidGar'
        roles: []
        __calls:
            - setType: [<(App\Entity\Config::TYPE_DEM)>]
            - addTiers: ['@tiers_with_another_config']
```

theofidry:
A wild guess: the metadata is injected & cached somewhere resulting in the current restore not to work. I'll revert this PR and release a patch. We'll just go forward and deprecate this feature and remove it in the next major version.

I'm a bit under water atm so if I forget it this weekend don't hesitate to ping me next week

yakobe:
Any news on this? Would be good to get a patch.

We are currently stuck on https://github.com/theofidry/alicedatafixtures/releases/tag/1.0.1 because the changes to id, but we need https://github.com/theofidry/AliceDataFixtures/releases/tag/v1.1.2 to use symfony 5 because of https://github.com/theofidry/AliceDataFixtures/issues/123 😛 

Maybe it's possible to revert all the breaking changes to the id since 1.0.1 😉😬

theofidry:
>Maybe it's possible to revert all the breaking changes to the id since 1.0.1 😉😬

As far as I'm aware there is only one which is this PR. I'll gladly merge & tag a release reverting it

yakobe:
The problem is that this PR fixes https://github.com/theofidry/AliceDataFixtures/issues/105. That means if it is rolled back then we still have the problem in https://github.com/theofidry/AliceDataFixtures/issues/105.

theofidry:
But the problem described in #105 was existing from day one and I think the real fix is removing this hybrid approach.

Maybe though as an alternative to completely reverting #105 the restore could be done via a toggle, which would allow some like @soullivaneuh to still benefit from it; just not enabled by default since it appears to easily break things

yakobe:
I think we might be misunderstanding eachother a little 😄.

The bug in #105 was introduced in https://github.com/theofidry/alicedatafixtures/releases/tag/1.1.0. 
This pull request (#135) attempts to fix it but introduces a different bug. If we revert this pull request we still have the bug described in #105. In order to fix #105 we would need to revert the original code added in https://github.com/theofidry/alicedatafixtures/releases/tag/1.1.0 that caused it.

Maybe the best solution is to just take away the new feature of being able to manually assign id's that are auto-generated by a database and not do any manipulation of the schema. That would be the preference since it can very easily cause nasty side-effects. Let the database generate the id as it is meant to. We have other ways to handle id references if we need 🙂

theofidry:
>Maybe the best solution is to just take away the new feature of being able to manually assign id's that are auto-generated by a database and not do any manipulation of the schema.

Just to be clear: this is not a new feature, this is from day 1 because was already a feature in Alice 2.x.

It's a chicken egg problem, #105 fixes an issue of the metadata not being restored but yet restoring it after a flush is creating a different set of issues. Hence IMO:

- short term; make the fix introduced in #105 toggable (but is not the default)
- long term: deprecate the feature and remove it in 2.x

yakobe:
No, sorry. #105 is an issue that reports a bug introduced on v1.1.0. 

#135 is the PR that tries to fix it but introduces a new bug.

PR #145, opened 2020-08-23
Fix PHPUnit deprecations

theofidry:


jmsche:
@theofidry I can't reproduce this locally, but I think setting min. phpunit version to 9.3 should be enough to fix the coverage test :)

theofidry:
Closed by #150

PR #152, opened 2020-11-25
Don't disable the id generator if the entity has already been persisted

### LINKED ISSUE #151: Still has `missing an assigned ID for field  'id'` bug
```
"Entity of type App\Component\Media\Model\Asset is missing an assigned ID for field  'id'. The identifier generation strategy for this entity requires the ID field to be populated before EntityManager#persist() is called. If you want automatically generated identifiers instead you need to adjust the metadata mapping accordingly."
```

this bug is supposed to be fixed at [#135](https://github.com/theofidry/AliceDataFixtures/pull/135)

Version
```
autelook/alice-bundle                          2.7.3              Symfony bundle to manage fixtures with Alice and Faker.
nelmio/alice                                    3.7.1              Expressive fixtures generator
theofidry/alice-data-fixtures                   1.3.0              Nelmio alice extension to persist the loaded fixtures.
```

```
  <entity name="Xiaolvmo\Component\Media\Model\AbstractAsset" table="app_media_asset">
        <id name="id" column="id" type="integer">
            <generator strategy="AUTO" />
        </id>

        <field name="path" column="path" type="string" nullable="false" />
</entity>
```

benja-M-1:
Should fix #151

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ before_install:
   - phpenv config-rm xdebug.ini || true
   - PHPUNIT_FLAGS='--stop-on-failure --verbose'
 
+  # TODO: Allow using Composer version 2 when will wikimedia/composer-merge-plugin support it too.
+  - composer self-update --1
+
 install:
   - perl -pi -e 's/^}$/,"provide":{"ext-mongo":"*"}}/' composer.json
   - composer update --prefer-dist $COMPOSER_FLAGS
--- a/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
+++ b/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
@@ -58,17 +58,19 @@ class ObjectManagerPersister implements PersisterInterface
         if (isset($this->persistableClasses[$class])) {
             $metadata = $this->objectManager->getClassMetadata($class);
 
-            // Check if the ID is explicitly set by the user. To avoid the ID to be overridden by the ID generator
-            // registered, we disable it for that specific object.
-            if ($metadata instanceof ORMClassMetadataInfo) {
-                if ($metadata->usesIdGenerator() && false === empty($metadata->getIdentifierValues($object))) {
-                    $metadata = $this->configureIdGenerator($metadata);
+            if (!$this->objectManager->contains($object)) {
+                // Unless the object is new, check if the ID is explicitly set by the user. To avoid the ID to be
+                // overridden by the ID generator registered, we disable it for that specific object.
+                if ($metadata instanceof ORMClassMetadataInfo) {
+                    if ($metadata->usesIdGenerator() && false === empty($metadata->getIdentifierValues($object))) {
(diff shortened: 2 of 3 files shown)


benja-M-1:
The build is failing because vendor-bin must be installed with composer 1 :/

benja-M-1:
The package wikimedia/composer-merge-plugin is not supporting Composer 2 yet. There is a branch that waits for tests, see https://github.com/wikimedia/composer-merge-plugin/issues/184 but for now AliceDataFixtures must rely on Composer 1 for the development environment.

benja-M-1:
YaY everything pass! :) @theofidry do you need something else for the review? Let me know if I can do something more.

theofidry:
Thank you @benja-M-1! Sorry for the delay I've just been a bit short of time

benja-M-1:
No pb I understand you have other things to do than open source :) Thank you for the work you provided until now on this package and the others you maintain or are involved in.

VincentLanglet:
Hi @benja-M-1 @theofidry, all my tests are failing since this "bugFix".
And I never had any issues related to #151, so I don't think this PR is the right way to solve it.

I do want to control all the id of my fixtures.

In my functional tests, I can look for an url `/foo/bar/3`.
In my repository tests, I can look for the entity with an id in (3, 4, 5).

I don't want my tests to start failing because I introduced a new fixture which changed the order of the persisted Entity (and their ID of course).

Now, the following fixtures
```
App\Entity\Foo
    test1:
        id: 1
        name: test1
    test2:
        id: 2
        name: test2
    test3:
        id: 3
        name: test3
```
Is imported in my database this way
```
test1 (id:1) 
test3 (id:2) 
test2 (id:3) 
```

theofidry:
Well that sucks I hope we can fix it. IMO I think I’ll try to get 2.x out
soon

benja-M-1:
Hi @VincentLanglet, in your project do your entities have relations to other? 
The scenario you provide seams to be already covered by [this test](https://github.com/theofidry/AliceDataFixtures/blob/master/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php#L100). Is it possible for you to share the fixtures?

VincentLanglet:
> Hi @VincentLanglet sorry to here that :(

Happens ;)

> In your project do your entities have relations to other?
> The scenario you provide seams to be already covered by [this test](https://github.com/theofidry/AliceDataFixtures/blob/master/tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php#L100). Is it possible for you to share the fixtures?

Yes, they have relations to other.
I provide a better example in the issue: https://github.com/theofidry/AliceDataFixtures/issues/153#issue-755156878

I assume that my fixture are persisted because of some others fixtures.
But then, when I want to hydrate the fixture the id I ask to be set is not set, since the object manager contains already the fixture.

I can't share the fixture since the project is private. 
I hope all the description I gave will be enough to understand and fix the issue ; if not I'll try to find extra time to create a minimal repository.

Issue #153, opened 2020-12-02
Since 1.3.1, fixture doesn't respect the provided ID.


VincentLanglet:
Before 1.3.1 version, the following fixtures
```
App\Entity\Bar
    bar1:
        id: 1
        foo: @foo1
    bar2:
        id: 2
        foo: @foo3
App\Entity\Foo
    foo1:
        id: 1
        name: foo1
    foo2:
        id: 2
        name: foo2
    foo3:
        id: 3
        name: foo3
```
Were correctly imported, with the provided ID

Since https://github.com/theofidry/AliceDataFixtures/pull/152, the ID provided are not respected anymore
- foo3 has the ID 2 since it's persisted by bar2
- foo2 has the ID 3 then

Firstly, this make failing a lot of my tests so this shouldn't be considered as a bugFix, more like a bug or a BC-break.
Secondly, I think the ID provided should always have the priority ; so I think that #152 is not the right fix of the #151 issue.

benja-M-1:
Which driver do you use? Mysql or PGSQL? As the generator does not behave the same it can have some importance.

VincentLanglet:
> Which driver do you use? Mysql or PGSQL? As the generator does not behave the same it can have some importance.

MariaDB, so it's Mysql

benja-M-1:
I created a test case here that fails with the reasons you said: https://github.com/benja-M-1/AliceDataFixtures/tree/fix/id

benja-M-1:
Ok this time the test fails to the good reason!

benja-M-1:
@theofidry After digging a bit on the issue I think that the PR #152 must be reverted.

Initially this PR tried to address an issue that happened if there were a call to the entityManager::flush method before the fixtures are loaded or in a listener triggered by the the doctrine event manager (which is actually a bad practice). 
Unfortunately this creates the bug @VincentLanglet is experiencing.

There is no difference from Doctrine point of view between an entity having an id set by a user or having an id set after an insert when it is persisted. For now I don't see how to manage the two cases cleanly. 

The only solution I can think of is that
- the user sets an id for every entities in its fixtures
- the user must clear the entity manager before loading the fixtures

I am very sorry if others have the issue :/

yakobe:
Interesting scenario. I'll try to give it a test if clearing the entity manager before loading fixtures solves the problem I hit in https://github.com/theofidry/AliceDataFixtures/issues/151#issuecomment-732819507.

[...]

PR #155, opened 2020-12-04
Keep the ids specified for entities by end users

benja-M-1:
This should solve the problem found in #153 
It also allows to mix fixtures with and without provided ids.

@VincentLanglet would you mind to test this PR with your fixtures set?
And @yakobe could you test with yours?

I checked on my project and this works.

### DIFF
--- a/fixtures/Bridge/Doctrine/Entity/DummyWithRelation.php
+++ b/fixtures/Bridge/Doctrine/Entity/DummyWithRelation.php
@@ -27,7 +27,7 @@ class DummyWithRelation
 
     /**
      * @var DummyWithIdentifier
-     * @OneToOne(targetEntity="Dummy", cascade={"persist"})
+     * @OneToOne(targetEntity="DummyWithIdentifier", cascade={"persist"})
      * @JoinColumn(name="dummy_id", referencedColumnName="id")
      */
     public $dummy;
--- a/fixtures/Bridge/Doctrine/Entity/DummyWithRelation.php
+++ b/fixtures/Bridge/Doctrine/Entity/DummyWithRelation.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Fidry\AliceDataFixtures package.
+ *
+ * (c) Théo FIDRY <theo.fidry@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Fidry\AliceDataFixtures\Bridge\Doctrine\Entity;
+
+/**
+ * @Entity
+ */
+class DummyWithRelation
+{
+    /**
+     * @Id
+     * @Column(type="integer")
+     * @GeneratedValue
+     */
+    public $id;
+
+    /**
+     * @var DummyWithIdentifier
+     * @OneToOne(targetEntity="Dummy", cascade={"persist"})
+     * @JoinColumn(name="dummy_id", referencedColumnName="id")
+     */
+    public $dummy;
+}
--- a/src/Bridge/Doctrine/IdGenerator.php
+++ b/src/Bridge/Doctrine/IdGenerator.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Fidry\AliceDataFixtures\Bridge\Doctrine;
+
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Id\AbstractIdGenerator;
+
+class IdGenerator extends AbstractIdGenerator
+{
+    const GENERATOR_TYPE_ALICE = 10;
+    /**
+     * @var AbstractIdGenerator
+     */
+    private $decorated;
+
+    public function __construct(AbstractIdGenerator $decorated)
+    {
+        $this->decorated = $decorated;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function generate(EntityManager $em, $entity)
+    {
+        $class = get_class($entity);
+
+        $metadata = $em->getClassMetadata($class);
+        $idValues = $metadata->getIdentifierValues($entity);
+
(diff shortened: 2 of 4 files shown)


VincentLanglet:
> @VincentLanglet would you mind to test this PR with your fixtures set?

It seems to fix my issue #153

yakobe:
Firstly, thank you @benja-M-1 for trying to find a solution that satisfies all of these different situations 👍.

Unfortunately this does not work for me yet. It causes a weird error like this: 

```
In AbstractMySQLDriver.php line 128:

  [Doctrine\DBAL\Exception\DriverException]
  An exception occurred while executing 'INSERT INTO user_group (name, description, roles) VALUES (?, ?, ?)' with params [null, "Demo User Group Name", "Demo Description", "[]"]:

  SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens
```

I'm trying to investigate what the cause could be. Maybe it has something to do with the fact that we have multiple "waves" of fixtures that run something like this (simplified for readability):

```php
$initialObjects = $this->fixtureLoader->load($initialFixtureFiles);
$this->fixtureLoader->load(
    $demoFixtureFiles,
    [],
    $initialObjects, // <-- these initial objects were created in the first wave above.
    PurgeMode::createNoPurgeMode()
);

```

@benja-M-1 any ideas how it could be causing this error?

benja-M-1:
@yakobe Do you use composable identifier for the user_group model?

VincentLanglet:
Hi

Reading this https://github.com/theofidry/AliceDataFixtures/issues/151#issuecomment-732819507
I feel like @yakobe  already had issues with this library before the first @benja-M-1 PR

So this PR currently doesn't solve everything but seems better than the 1.3.0 version AND fix the regression introduced in 1.3.1. It could be a good first step.

Anyway something should be done to fix the current solution @theofidry :
- either reverting https://github.com/theofidry/AliceDataFixtures/pull/152
- either merging this

theofidry:
Let's hope it does it then, thank you @benja-M-1

VincentLanglet:
Thanks @theofidry ; do you mind to create a release 1.3.2 ? :)

theofidry:
I would like to get the CI green first... I merged this one because the failure is unrelated

yakobe:
@VincentLanglet I understand that you want it to work for you, but this change just swaps who it does not work for (i.e. now it works for you and not for me). 

It would be nicer if we could have worked together to find a solution that would work for all and in the meantime we can "fix" our versions to a specific release that works for us. That is what I will need to do until we can find such a solution. Not critical, just a little unfair.

VincentLanglet:
> @VincentLanglet I understand that you want it to work for you, but this change just swaps who it does not work for (i.e. now it works for you and not for me).

I want this to work at least as good as it was at the 1.3.0.
I proposed two solutions: merging this or reverting the original PR.

The 1.3.1. introduce a BC-break. First thing to do is to fix the BC-break ; there is nothing related to me or you.

> It would be nicer if we could have worked together to find a solution that would work for all and in the meantime we can "fix" our versions to a specific release that works for us. That is what I will need to do until we can find such a solution. Not critical, just a little unfair.

You're welcome to answer to benja-M-1 questions in order to solve the situation. @yakobe 
> Do you use composite identifier for the user_group model?
> I may need your model to dig what could cause the issue. Can you make a public github repo are post here the model (with relations if any)?

yakobe:
No need to be condescending. As I said, I understood your motives, it just would have been nicer to find a solution that favors us all.

> I want this to work at least as good as it was at the 1.3.0.

Totally understand that. We all want it to work.

> First thing to do is to fix the BC-break

Yeah... it was fixing the one introduced in 1.1.0. Now we just have a different BC break.

> You're welcome to answer to benja-M-1 questions in order to solve the situation. @yakobe

I was getting to it. Now this is merged it will need to go on a new issue rather than here.

yakobe:
@benja-M-1 Sorry for the delay. The `user_group` model has a normal generated primary key:

```
    /**
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private ?int $id = null;
```

Not sure if that helps. I think it happens because our fixtures use existing db objects like i mention [above](https://github.com/theofidry/AliceDataFixtures/pull/155#issuecomment-740219186). I guess we are maybe using that feature differently from it's original intention and therefore I'm going to see if i can re-arrange our fixtures so that it's not needed. With a bit of luck it should be possible and we can live with the change.

benja-M-1:
@yakobe Would it be possible to setup a repo that contains enough code to
reproduce the bug? Or even better providing a failing test case?

I would be glad to help as much as I can.

Le lun. 14 déc. 2020 à 17:17, Jake Bishop <notifications@github.com> a
écrit :

yakobe:
@benja-M-1 thanks for the follow up. I really appreciate your will to help.

Our repo is a bit of a monster and as with most closed source projects i cannot reveal too much 😕. Setting up a new project might take a while because i'm not sure familiar with all the components. 

You dont know of a "basic" project that is used for testing this repo do you? If so, i could take that as starting point and attempt to break it in a similar way 😅.

yakobe:
@benja-M-1 fyi i have managed to re-structure our tests so that this problem no-longer occurs for me. Hopefully we were the only ones with this weird side-effect and therefore a fix doesn't need any priority 😺 .

VincentLanglet:
> @benja-M-1 fyi i have managed to re-structure our tests so that this problem no-longer occurs for me. Hopefully we were the only ones with this weird side-effect and therefore a fix doesn't need any priority 😺 .

Good news 😁

benja-M-1:
This is a great news! Glad to read that :) hopefully nobody will complain
about this issue anymore 😅


Le mar. 15 déc. 2020 à 18:14, Vincent Langlet <notifications@github.com> a
écrit :

PR #156, opened 2020-12-04
Use an env var to change which composer executable used by make

benja-M-1:
As the package can't be upgraded to Composer 2 and must use Composer 1 it is useful to specify which binary of composer to use to run the test when developing.

For example:
```
COMPOSER_BIN="php -d memory_limit=-1 composer.phar" make test
```


alexislefebvre:
You can update the PR or rebase it on master.

theofidry:
Sorry for being able to give attention to this PR in time. Now that the library only supports Composer 2 I believe this is no longer necessary. Thanks for the PR nonetheless

PR #158, opened 2020-12-09
Support PHP 8.0

alexislefebvre:
Let's see what happens.


alexislefebvre:
It will require an update from https://github.com/nelmio/alice

See https://github.com/nelmio/alice/pull/1064

nicodmf:
@theofidryany news on this ?

alexislefebvre:
This PR is blocked by a PR that is blocked by another PR. :sweat_smile:

Yozhef:
@alexislefebvre nelmio/alice#1064 is done =)

alexislefebvre:
It is blocked by https://github.com/theofidry/composer-inheritance-plugin/pull/8 now.

alexislefebvre:
https://github.com/doctrine/mongodb-odm doesn't support PHP 8 yet.

VincentLanglet:
Can't we skip temporary the mongo-related test on php8 ?
Doctrine/orm supports php8, it should be allowed to use this package on php8 if we don't use mongo.

alexislefebvre:
I thought about that, and I think it would be confusing to publish a new release that may break users' tests (if they didn't locked their dependencies, that's a good practice but some users don't do that).

VincentLanglet:
> I thought about that, and I think it would be confusing to publish a new release that may break users' tests (if they didn't locked their dependencies, that's a good practice but some users don't do that).

Why will it break the users tests ?
You're just allowing php8 in this PR.

User without mongo, will be able to upgrade to PHP8, user with mongo won't be able to upgrade to PHP8 ; not because of us, but because of mongo which doesn't currently support PHP8.

There is no change in the code.
If someone use this library on PHP7.4 with mongo, it will still work the same ; proof => test are green on 7.4 with mongo.

alexislefebvre:
@VincentLanglet you're totally right, I overthought that. :slightly_smiling_face:

stof:
The composer-inheritance-plugin is now solved, so this work should be resumed

.github/workflows/tests.yaml
```diff
@@ -61,5 +62,11 @@ jobs:
                     dependency-versions: ${{ matrix.dependencies }}
                     composer-options: "--prefer-dist --prefer-stable"
 
+            -   name: Remove dependencies not compatible with PHP 8
```
VincentLanglet, line 65: I think you should create a make test/php8 command instead in order to not run the mongo tests.
https://github.com/theofidry/AliceDataFixtures/blob/master/Makefile#L116

alexislefebvre: Thanks for your help but I didn't want to duplicate all of these lines: https://github.com/theofidry/AliceDataFixtures/blob/22040ef893953a74f02f7c678cb2f14641cb7d60/Makefile#L80

And [`make test_symfony_doctrine_bridge_proxy_manager`](https://github.com/theofidry/AliceDataFixtures/blob/22040ef893953a74f02f7c678cb2f14641cb7d60/Makefile#L184) will call that [configuration](https://github.com/theofidry/AliceDataFixtures/blob/7165a10e4e794852ea50d9783873f7ee34f68516/phpunit_symfony_proxy_manager_with_doctrine.xml.dist#L28) and that will launch [several tests](https://github.com/theofidry/AliceDataFixtures/tree/f2c9e5432c9e0a61f2d4d6440c295a412951945e/tests/Bridge/Symfony/Doctrine), and only one require `doctrine/mongodb-odm-bundle`.

So I thought it was simpler to remove the dependency and skip the test that rewrite Makefile or PHPUnit configuration.

alexislefebvre:
One job looks dead but the PHP8 job is green. Feel free to review the changes.

theofidry:
Thank you!

.github/workflows/tests.yaml
alexislefebvre: It will be addressed in #168

PR #167, opened 2021-01-18
Compatible with PHP 8.0

olvlvl:


alexislefebvre:
The upgrade to allow PHP 8.0 is blocked by some dependencies: #158

PR #174, opened 2021-05-13
Fix MySQL platform detection

colinodell:
When purging data, this library attempts to detect whether MySQL is used by checking the "driver"; if so, it temporarily disables foreign key constraints.

However, some libraries (like Sentry's Symfony bundle) decorate that driver to add functionality like tracing the SQL statements being executed.  Because these decorators do not extend from `AbstractMySQLDriver` this library fails to disable foreign key checks when decorated drivers are being used.  See https://github.com/getsentry/sentry-symfony/issues/503 for an example of this.

I propose that this library should check the **platform** instead of the **driver**.  Platforms describe **what type of database** is being used, versus the drivers which provide **how to interface with that database**.  Because we care more about the "what" and not the how, and because platforms are far less likely to be overridden/decorated, this change should avoid compatibility issues with other libraries and custom code.

### DIFF
--- a/src/Bridge/Doctrine/Purger/Purger.php
+++ b/src/Bridge/Doctrine/Purger/Purger.php
@@ -17,7 +17,7 @@ use Doctrine\Common\DataFixtures\Purger\MongoDBPurger as DoctrineMongoDBPurger;
 use Doctrine\Common\DataFixtures\Purger\ORMPurger as DoctrineOrmPurger;
 use Doctrine\Common\DataFixtures\Purger\PHPCRPurger as DoctrinePhpCrPurger;
 use Doctrine\Common\DataFixtures\Purger\PurgerInterface as DoctrinePurgerInterface;
-use Doctrine\DBAL\Driver\AbstractMySQLDriver;
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\ODM\MongoDB\DocumentManager as DoctrineMongoDocumentManager;
 use Doctrine\ODM\PHPCR\DocumentManager as DoctrinePhpCrDocumentManager;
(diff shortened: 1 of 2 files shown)


theofidry (approved):

ste93cry:
What's the status of this?

epitre:
Hi @colinodell 
How did you manage to work around it without this pr merged ?

@theofidry Do you think there's a lot to do to have it released ?

Bests

colinodell:
We're using cweagans/composer-patches to apply this PR's diff (add `.diff` to the URL)

theofidry (approved):

theofidry:
I am not sure why the build gets cancelled :/ The patch looks good to me though

epitre:
> I am not sure why the build gets cancelled :/ The patch looks good to me though

can you relaunch the builds, so you can merge it ?

epitre:
> I am not sure why the build gets cancelled :/ The patch looks good to me though

maybe that's why : 
```
PHP Fatal error:  Uncaught RuntimeException: Setup tool cannot configure caches without doctrine/cache 1.11 or symfony/cache. Please add an explicit dependency to either library. in /home/runner/work/AliceDataFixtures/AliceDataFixtures/vendor-bin/doctrine/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/Setup.php:184
```

epitre:
it seems that the problem is in master :/
https://github.com/theofidry/AliceDataFixtures/runs/3713999940

colinodell:
This is caused by a minor BC break in Doctrine: https://github.com/doctrine/orm/blob/2.9.x/UPGRADE.md#minor-bc-break-setup-tool-needs-cache-implementation

PR #201, opened 2021-12-21
Fix persisting entity with nullable ManyToOne relation

javer:
Fixes the error introduced in #198 (and release 1.5.0) during persisting an entity with a nullable ManyToOne relation:
```
TypeError: Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister::getMetadata(): Argument #2 ($object) must be of type object, null given, called in src/Bridge/Doctrine/Persister/ObjectManagerPersister.php on line 158
```

See failed build with the test without fix: https://github.com/javer/AliceDataFixtures/actions/runs/1607767589
And successful build with the test and fix: https://github.com/javer/AliceDataFixtures/actions/runs/1607777714

### DIFF
--- a/fixtures/Bridge/Doctrine/Entity/DummyWithRelatedCascadePersist.php
+++ b/fixtures/Bridge/Doctrine/Entity/DummyWithRelatedCascadePersist.php
@@ -41,6 +41,11 @@ class DummyWithRelatedCascadePersist
     public Dummy $related;
 
     /**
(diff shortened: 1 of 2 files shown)


akalineskou:
Yup, I'm having the same issue... tried the same fix and it works!

theofidry:
Thank you @javer, released in https://github.com/theofidry/AliceDataFixtures/releases/tag/1.5.1

Kynno:
Thank you so much for the quick fix!

Issue #202, opened 2022-01-03
ObjectManagerPersister bad arguments


Helo3615:
Hello

Following this issue : https://github.com/theofidry/AliceBundle/issues/6

I just updated `theofidry/alice-data-fixtures` to 1.5.1 and got this error by launching fixtures :
```
  Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister::g  
  etMetadata(): Argument #2 ($object) must be of type object, array given, ca  
  lled in /app/vendor/theofidry/alice-data-fixtures/src/Bridge/Doctrine/Persi  
  ster/ObjectManagerPersister.php on line 158  
```

I updated `hautelook/alice-bundle` to 2.10 as well.

Thanks you

fkizewski:
Hello

I've the same issue since i've updated `theofidry/alice-data-fixtures` to 1.5.1.

I my case, it's appear because into my fixtures i'm using array:
```
App\Entity\User:
  user_2:
    roles: ['@role_*']
```

To testing and try to understand where is the problem, i've added this condition into the ObjectManagerPersister before the testing line with null, and all is fine now (i known it's a bad idea to do this) :
```
elseif (is_array($fieldValueOrFieldValues)) {
    foreach ($fieldValueOrFieldValues as $fieldValue) {
        $this->getMetadata($targetEntityClassName, $fieldValue);
    }
}
```
There is another possibility to replace an array by a collection into fixtures (because collection is managed)? Or it's necessary to update the library?

Thanks for your help.

theofidry:
Would it be possible to have access to a reproducer or a more complete sample of the issue? i.e. the fixture, the model behind and the complete stack trace? Because with the current information I have no idea what is causing the issue neither where is it coming from

fkizewski:
Thanks @theofidry for your reply.
Please find a more complete sample :

**Entities**
User.php
```
<?php

namespace App\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity()
 */
class User
{
    /**
     * @ORM\Id()
     * @ORM\GeneratedValue()
     * @ORM\Column(type="integer")
     */
    private ?int $id;

    /**
     * @var ArrayCollection<int, Group>
     *
     * @ORM\ManyToMany(targetEntity="App\Entity\Group", inversedBy="users")
     */
    public $groups;

    public function __construct()
    {
[...]

PR #203, opened 2022-02-02
fix(persister): added condition to check if entity values are array

santoslluis:
I had the same issue here: https://github.com/theofidry/AliceDataFixtures/issues/202
It was working well until I update the library.

## Error context:
```
TypeError : Argument 2 passed to Fidry\AliceDataFixtures\Bridge\Doctrine\Persister\ObjectManagerPersister::getMetadata() must be an object, array given
```

I defined the relationship datatype as array and not as Collection or ArrayCollection.

```php
    /**
     * @var Instance[]
     * @ORM\OneToMany(targetEntity=Instance::class, mappedBy="strategy")
     */
    public $instances = [];
```

It seems that if you didn't specify the datatype as Collection or ArrayCollection, the library loads it as array, and after in inner checks It only checks if it's a Collection, otherwise threats it as a Object, so it throws a `TypeError`.

## What I did
I added a check if it is an array, and processed it like if it was a Collection.

For me it is working, hope it helps.

### DIFF
--- a/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
+++ b/src/Bridge/Doctrine/Persister/ObjectManagerPersister.php
@@ -154,6 +154,10 @@ class ObjectManagerPersister implements PersisterInterface
                 foreach ($fieldValueOrFieldValues->getValues() as $fieldValue) {
                     $this->getMetadata($targetEntityClassName, $fieldValue);
                 }
+            } elseif (is_array($fieldValueOrFieldValues)){
+                foreach ($fieldValueOrFieldValues as $fieldValue) {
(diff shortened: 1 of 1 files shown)


theofidry:
Thanks for the PR! Would it be possible to have an anti-regression test as well? It is a sensitive part which has broke quite a few times already... I think you should find a couple of examples in the test suite already and running the tests should only be a matter of doing `make test` (hopefully). If you need more help there don't hesitate to ping me.

theofidry:
Sorry this is long overdue but I couldn't reproduce it in #282, hence I'll be closing this down.

PR #216, opened 2022-07-01
use a class for read values from env

alexislefebvre:
The doc of composer doesn't mention the support of `files`: https://getcomposer.org/doc/04-schema.md#autoload-dev

It works by using a class.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,10 +68,7 @@
                 "fixtures",
                 "tests"
             ]
-        },
-        "files": [
-            "fixtures/get-param.php"
-        ]
+        }
     },
 
     "config": {
--- a/doctrine-odm-db-settings.php
+++ b/doctrine-odm-db-settings.php
@@ -2,10 +2,12 @@
 
 declare(strict_types=1);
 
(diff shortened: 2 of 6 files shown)


theofidry:
It may be missing from the doc but it is there:

```
"files": {
                    "type": "array",
                    "description": "This is an array of files that are always required on every request."
                }
```

See the `composer-schema.json` file. You can also see the result in the dumped autoloading files (`vendor/composer/autoload_files.php` for this one)

It does require the autoloader to be dumped for it to be taken into account though.

alexislefebvre:
I still have the issue after running `composer dump-autoload` on my local environment. And the CI is red with the same error.

Maybe the issue is that the path is not correct when reading `composer.json` from a subdirectory.

alexislefebvre:
It occurred in https://github.com/theofidry/AliceDataFixtures/runs/7156925224?check_suite_focus=true

But you fixed it in the commit after, now the error is about Monolog: https://github.com/theofidry/AliceDataFixtures/runs/7157015613?check_suite_focus=true

PR #245, opened 2023-12-02
ci: Migrate from mongo to mongosh

theofidry:
Starting from Mongodb version 6.0 mongo was replaced by mongosh


theofidry:
Thanks @ghostwriter

PR #264, opened 2024-02-07
fix: clear entity persister on restoring class metadata

wickedOne:
restoring the class metadata in the metadata factory does not restore the one doctrine has cached in the `class` property of it's persisters....

downstream this might lead to unexpected behaviour: when the id generator has been replaced with this library's `IdGenerator`, the basic entity persister does not recognise it as an id generator and treats the identity field as one which can be used for inserts in the [`BasicEntityPersister::getInsertColumnList`](https://github.com/doctrine/orm/blob/94144e122779098c156098fc9bb6791e53192086/src/Persisters/Entity/BasicEntityPersister.php#L1443-L1451) method 

this can lead to `Invalid parameter number: number of bound variables does not match number of tokens` errors on subsequent inserts of the entity manager. this change will fix that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,7 +221,6 @@ vendor-bin/doctrine/composer.lock: vendor-bin/doctrine/composer.json
 	@echo vendor-bin/doctrine/composer.lock is not up to date.
 
 vendor-bin/doctrine/vendor/phpunit: vendor-bin/doctrine/composer.lock
-	composer bin doctrine update $(COMPOSER_FLAGS) || true
 	composer bin doctrine update $(COMPOSER_FLAGS)
 	touch $@
 
--- a/Makefile
+++ b/Makefile
@@ -221,6 +221,7 @@ vendor-bin/doctrine/composer.lock: vendor-bin/doctrine/composer.json
 	@echo vendor-bin/doctrine/composer.lock is not up to date.
 
 vendor-bin/doctrine/vendor/phpunit: vendor-bin/doctrine/composer.lock
+	composer bin doctrine update $(COMPOSER_FLAGS) || true
 	composer bin doctrine update $(COMPOSER_FLAGS)
 	touch $@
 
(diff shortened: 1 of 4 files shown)


theofidry:
Thanks for the PR @wickedOne, sorry for the late review. Would there be a way to capture this behaviour in the tests?

wickedOne:
no probs: better late than never :-)

as the persisters are not publicly available, i would require some hacky trickery. 
if that's ok with you, i can probably come up with something

theofidry:
does it require to touch the persister class though (the test)? From your PR description, I thought this was somewhat reproducible by loading some fixtures and then doing inserts

wickedOne:
ah ofc yes, the other way around :-)
yes, that _should_ be possible, i'll giv that a go next week

wickedOne:
added test which makes sure the entity manager doesn't crash after flushing the entity persister.
so basically remove my modification to the entity persister to see that test fail

the modification to the makefile is somewhat unrelated, but was necessary for me to run it locally (bin directory wasn't created in the vendor bin for doctrine)

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
```diff
@@ -140,6 +140,24 @@ public function testCanPersistMultipleEntitiesWithExplicitIdentifierSet(): void
...
 
+    public function testClearingUnitOfWorkPersister(): void
+    {
+        $dummy = new DummyWithProperty();
+        $dummy->id = 1;
+
+        $this->persister->persist($dummy);
+        $this->persister->flush();
+
+        $dummy = new DummyWithProperty();
```
theofidry, line 151: I think you're missing the import statement

Makefile
```diff
@@ -221,6 +221,7 @@ vendor-bin/doctrine/composer.lock: vendor-bin/doctrine/composer.json
 	@echo vendor-bin/doctrine/composer.lock is not up to date.
 
 vendor-bin/doctrine/vendor/phpunit: vendor-bin/doctrine/composer.lock
+	composer bin doctrine update $(COMPOSER_FLAGS) || true
```
theofidry, line 224: should be reverted

theofidry (approved):
Two tiny things to fix LGTM otherwise, thanks a lot for that

tests/Bridge/Doctrine/Persister/ObjectManagerPersisterTest.php
wickedOne: done

Makefile
wickedOne: done

theofidry:
I checked the failures and they look unrelated, thank you @wickedOne

PR #267, opened 2024-09-17
fix purge without foreign check on mariadb (doctrine/dbal:^4.0)

gassan:
Since `doctrine/dbal:4.0`, `MariaDBPlatform` no longer extends from `MySqlPlatform`, so you cannot truncate tables because `foreign_key_checks` are not disabled.

### DIFF
--- a/src/Bridge/Doctrine/Purger/Purger.php
+++ b/src/Bridge/Doctrine/Purger/Purger.php
@@ -17,6 +17,7 @@ use Doctrine\Common\DataFixtures\Purger\MongoDBPurger as DoctrineMongoDBPurger;
 use Doctrine\Common\DataFixtures\Purger\ORMPurger as DoctrineOrmPurger;
 use Doctrine\Common\DataFixtures\Purger\PHPCRPurger as DoctrinePhpCrPurger;
 use Doctrine\Common\DataFixtures\Purger\PurgerInterface as DoctrinePurgerInterface;
+use Doctrine\DBAL\Platforms\MariaDBPlatform;
 use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\ODM\MongoDB\DocumentManager as DoctrineMongoDocumentManager;
 use Doctrine\ODM\PHPCR\DocumentManager as DoctrinePhpCrDocumentManager;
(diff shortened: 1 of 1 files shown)


src/Bridge/Doctrine/Purger/Purger.php
```diff
@@ -90,10 +91,14 @@ public function purge(): void
         // PostgreSQL. This ideally should be done in the Purger of doctrine/data-fixtures but meanwhile we are doing
         // it here.
         // See the progress in https://github.com/doctrine/data-fixtures/pull/272
+        $platform = $this->purger->getObjectManager()->getConnection()->getDatabasePlatform();
         $disableFkChecks = (
             $this->purger instanceof DoctrineOrmPurger
             && in_array($this->purgeMode->getValue(), [PurgeMode::createDeleteMode()->getValue(), PurgeMode::createTruncateMode()->getValue()])
-            && $this->purger->getObjectManager()->getConnection()->getDatabasePlatform() instanceof MySqlPlatform
+            && (
```
theofidry, line 98: let's move this into a method instead:

```php
private function doesDatabaseSupportsTruncate(): bool {
  $platform = $this->purger->getObjectManager()->getConnection()->getDatabasePlatform();

  return $platform instanceof MariaDBPlatform
                || $platform instanceof MySqlPlatform;
}
```

theofidry (approved):
👍 thanks for the PR!

src/Bridge/Doctrine/Purger/Purger.php
gassan: is done.

src/Bridge/Doctrine/Purger/Purger.php
```diff
@@ -136,4 +137,12 @@ private static function createPurger(ObjectManager $manager, ?PurgeMode $purgeMo
...
         );
     }
+
+    private function doesDatabaseSupportTruncate(): bool
+    {
+        $platform = $this->purger->getObjectManager()->getConnection()->getDatabasePlatform();
+
+        return $platform instanceof MySqlPlatform
+            || $platform instanceof MariaDBPlatform;
+    }
```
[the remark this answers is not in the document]
gassan: You do not need this complex check.
MariaDBPatform extends in past from MySQLPlatform

PR #283, opened 2024-12-08
Fix code base for support php 8.4

sonrac:
https://www.php.net/ChangeLog-8.php
https://www.php.net/releases/8.4/en.php

`Implicitly nullable parameter types are now deprecated.`

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -24,6 +24,7 @@ jobs:
             matrix:
                 php-version:
                     - '8.3'
+                    - '8.4'
                 dependencies: [ highest, lowest ]
                 make-test:
                     - test_core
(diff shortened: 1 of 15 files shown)


phansys (commented):
The rule `'nullable_type_declaration_for_default_null_value' => true` SHOULD be added to the PHP CS Fixer rules.

https://github.com/theofidry/AliceDataFixtures/blob/53ff37e432852b77d5eb53a330b74d04c74cb62b/.php-cs-fixer.dist.php#L38-L91

themasch:
@sonrac thanks for starting on this. Do you want to pick up that php-cs-fixer rule change @phansys requested, or would it be okay for you if I take your branch, add the cs-fixer rule and create a new PR so we can get this merged?

PR #287, opened 2025-03-05
Fix for doctrine 3.0

drewblin:
Docrine ORM 3.0 has such BC-break changes:
* Signature of AbstractIdGenerator changed
* ClassMetadataInfo was removed
* $metadata->getAssociationMappings() returns object instead of array

Note, changes in this pull request are incompatible with Doctrine ORM 2.0. So, I'm not sure this corresponds to you vision of project. But you can use this PR to understand what code should be fixed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "theofidry/alice-data-fixtures",
+    "name": "drewblin/alice-data-fixtures",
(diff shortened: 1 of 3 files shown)


theofidry:
Thanks for the PR! I rebased it in #287.

theofidry:
Merged, thanks again!

PR #333, opened 2025-12-15
test: Fix the leaking error handler by installing Symfony runtime

theofidry:


theofidry:
@xabbuh this still makes no sense to me... Would Symfony be open to provide the API to unregister the `ErrorHandler` and do so in the `KernelTestCase`?

xabbuh:
@theofidry What is the issue without the Runtime component being installed?

theofidry:
@xabbuh the issue reported in https://github.com/symfony/symfony/issues/53812. PHPUnit fails with:

```
Test code or tested code did not remove its own exception handlers
```

Having the Runtime component installed fixes the issue because then the error handler used is the one from the runtime which can and is unregistered on shutdown.

xabbuh:
Got it, I guess that's the same (IMO wrong) behaviour that's also discussed here: https://github.com/symfony/symfony/pull/58372#discussion_r1774862781 Can you confirm?

theofidry:
I can confirm. I was looking for that comment/piece of code but GitHub was not loading for me 😅

